PR #24, opened 2018-08-29
Store Vault connection parameters in etcd

zoetrope:
In addition, the default config path is changed.

### DIFF
--- a/cli/ca.go
+++ b/cli/ca.go
@@ -1,0 +1,111 @@
+package cli
+
+import (
+	"context"
+	"flag"
+	"fmt"
+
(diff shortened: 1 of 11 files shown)


cli/ca.go
```diff
@@ -0,0 +1,111 @@
...
+	return newc.Execute(ctx)
+}
+
+// CACommand implements "ca" subcommand
+func CACommand() subcommands.Command {
+	return subcmd{
+		ca{},
+		"ca",
+		"set the ca configuration",
+		"ca config JSON",
```
ymmt2005: ?

PR #43, opened 2018-09-12
Service Accounts

ymmt2005:
This PR enables k8s service accounts.

### DIFF
--- a/pki.go
+++ b/pki.go
@@ -2,7 +2,11 @@ package cke
 
 import (
 	"context"
+	"path"
 	"path/filepath"
+
+	"github.com/cybozu-go/log"
+	vault "github.com/hashicorp/vault/api"
 )
 
 // CA Keys in Vault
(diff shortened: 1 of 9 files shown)


pki.go
```diff
@@ -97,10 +148,34 @@ type KubernetesCA struct {
...
 // setup generates and installs certificates for API server.
 func (k KubernetesCA) setup(ctx context.Context, inf Infrastructure, node *Node) error {
-	err := writeCertificate(inf, node, CAKubernetes, K8sPKIPath("apiserver"),
+	sakey, err := inf.Storage().GetServiceAccountKey(ctx)
+	if err != nil {
+		return err
+	}
+	err = writeFile(inf, node, K8sPKIPath("service-account.key"), sakey)
+	if err != nil {
+		return nil
```
dulltz: Is it ok `return nil`?

ymmt2005: oops.  ur right.

PR #50, opened 2018-09-18
SELinux support

zoetrope:
* add `selinux` field to cluster configuration to each node and cluster.
 * add `propagation` and `mode` field to `Mount`. it apply to docker volume option.
 * set `shared` and `z` option to the shared directory among multiple containers


docs/cluster.md
```diff
@@ -10,6 +10,7 @@ Name            | Required | Type      | Description
 `name`          | true     | string    | The k8s cluster name.
 `nodes`         | true     | array     | `Node` list.
 `ssh_key`       | false    | string    | Cluster wide SSH private key.
+`selinux`       | false    | bool      | SELinux is enable on all nodes of the cluster.
```
ymmt2005: SELinux is enabled on all nodes; this overrides per-node `selinux` settings.

PR #64, opened 2018-09-25
Use machine-id to identify nodes

ueokande:
http://man7.org/linux/man-pages/man5/machine-id.5.html
> The machine ID is a single newline-terminated, hexadecimal, 32-character, lowercase ID.

The host name could be invalid machine id.  Mount host's `/etc/machine-id` to the container.


ymmt2005:
https://github.com/coreos/bugs/issues/2166

`machine-id` need to be specified externally by some means when
CoreOS Container Linux is booted by (i)PXE.  This should be addressed
in sabakan.

ymmt2005:
Anyway, this change looks good to me.

PR #73, opened 2018-10-13
Revamp mtest

ymmt2005:
mtest should not consider the strategy how CKE reconciles cluster with given conditions.
The strategy itself is tested extensively in [strategy_test.go](/cybozu-go/cke/tree/master/strategy_test.go).

Instead, mtest should cover all SSH commands invoked by CKE because they cannot be tested on a single host.  That can be achieved by running all operators.

mtest should also perform some smoke-testing for deployed Kubernetes cluster.


mtest/operators_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}).Should(Succeed())
+
+		By("Removing a control plane node from the cluster")
+		// this will run:
+		// - EtcdRemoveMemberOp
+		// - ToDo: K8sRemoveNodeOp
+		// - RiversRestartOp
+		// - APIServerRestartOp
+		stopCKE()
+		execAt(node2, "sudo", "-b", "reboot", "-f", "-f")
```
zoetrope: `-f` が2つあるのは何か意味がありますか？

ymmt2005: はい。man reboot says:

```
       -f, --force
           Force immediate halt, power-off, or reboot. When specified once, this results in an immediate but clean
           shutdown by the system manager. When specified twice, this results in an immediate shutdown without
           contacting the system manager. See the description of --force in systemctl(1) for more details.
```

zoetrope: なるほどです。 (debianの`man reboot`にはなかった…)

PR #81, opened 2018-10-24
Set deadline for SSH connection

ymmt2005:
golang.org/x/crypto/ssh has some known issues that block clients
indefinitely when SSH server dies.

Ref: golang/go#26643, golang/go#21420

To workaround the problem, this commit creates TCP connection
to the server by itself then passes it to `ssh.NewClientConn` to
control the underlying TCP connection directly.

It adds deadlines against the TCP connection before any SSH
activity.  It also enables TCP keepalive with short period.


agent.go
```diff
@@ -10,7 +11,21 @@ import (
...
+	defaultKeepAlive   = 5 * time.Second
+
+	// DefaultRunTimeout is the timeout value for Agent.Run().
+	DefaultRunTimeout = 10 * time.Minute
+)
+
+var (
+	// When KeepAlive is > 0, the dialer returns TCP connections
+	// with keep-alive enabled.  With the default 5 second duration,
+	// the implementation can detect dead peer in around 50 seconds.
```
zoetrope: `tcp_keepalive_probes` のデフォルト値が9回なので、50秒ということか。なるほど。

PR #127, opened 2019-02-06
[ckecli] ssh/scp command can use all nodes regardless cluster member

mitsutaka:
Background:
`ckecli ssh NODE` can use only node who is in the cluster member.
It is useless if user want to login to non-member node.

Expect:
`ckecli ssh` and `ckecli scp` can use all nodes regardless cluster
member.

### DIFF
--- a/pkg/ckecli/cmd/ssh.go
+++ b/pkg/ckecli/cmd/ssh.go
@@ -37,25 +37,7 @@ func writeToFifo(fifo string, data string) {
 	}
 }
 
-func resolveNode(ctx context.Context, nodeName string) (*cke.Node, error) {
-	cluster, err := storage.GetCluster(ctx)
-	if err != nil {
-		return nil, err
-	}
-	var node *cke.Node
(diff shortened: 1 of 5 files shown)


pkg/ckecli/cmd/ssh.go
```diff
@@ -102,12 +84,8 @@ func sshPrivateKey(node *cke.Node) (string, error) {
...
 
 func ssh(ctx context.Context, args []string) error {
-	nodeName := args[0]
-	node, err := resolveNode(ctx, nodeName)
-	if err != nil {
-		return err
-	}
-	fifo, err := sshPrivateKey(node)
+	sshAddress := args[0]
+	fifo, err := sshPrivateKey(sshAddress)
```
ymmt2005: `user@` がついているなら、外さないと sshPrivateKey のキーと
ノード名がマッチしません。

ymmt2005:
LGTM

PR #151, opened 2019-03-19
Fix user-defined resources operation

mitsutaka:
Service resource of the kubernetes can not update because ClsuterIP is already defined.

### DIFF
--- a/pkg/apply_gen/main.go
+++ b/pkg/apply_gen/main.go
@@ -36,7 +36,7 @@ func annotate(meta *metav1.ObjectMeta, rev int64, data []byte) {
 }
 {{- range . }}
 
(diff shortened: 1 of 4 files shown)


pkg/apply_gen/main.go
```diff
@@ -68,6 +65,9 @@ func apply{{ .Kind }}(o *{{ .API }}.{{ .Kind }}, data []byte, rev int64, getFunc
 	if err != nil {
 		return err
 	}
+	if !ok {
+		original = string(modified)
```
ymmt2005: `log.Warn()` で一応警告を出力しておきたいです。

pkg/apply_gen/main.go
```diff
@@ -68,6 +65,10 @@ func apply{{ .Kind }}(o *{{ .API }}.{{ .Kind }}, data []byte, rev int64, getFunc
...
+		log.Warn("use modified resource as original for 3-way patch", nil)
```
ymmt2005:
```suggestion
		log.Warn("use modified resource as original for 3-way patch", map[string]interface{}{
			"kind": o.Kind,
			"namespace": o.Namespace,
			"name": o.Name,
		})
```

PR #152, opened 2019-03-19
Do not schedule pods when the number of Ready nodes is less than half

dulltz:


status.go
```diff
@@ -36,14 +36,42 @@ type EtcdBackupStatus struct {
...
+			if cond.Type != corev1.NodeReady {
+				continue
+			}
+			if cond.Status == corev1.ConditionTrue {
+				currentReady++
+				break
+			}
+		}
+	}
+	if clusterNodesSize/2+1 <= currentReady {
```
ymmt2005: `return clusterNodesSize/2 < currentReady` でいけますね。

status.go
```diff
@@ -36,14 +36,41 @@ type EtcdBackupStatus struct {
...
+			if cond.Type != corev1.NodeReady {
+				continue
+			}
+			if cond.Status == corev1.ConditionTrue {
+				currentReady++
+				break
+			}
+		}
+	}
+	if clusterNodesSize/2 < currentReady {
```
ymmt2005: や、単純に `return clusterNodesSize/2 < currentReady` で。

PR #155, opened 2019-03-22
Fix merge method for user defined resource

ishii-masayuki:


pkg/apply_gen/main.go
```diff
@@ -107,22 +125,23 @@ func subMain() error {
 	defer f.Close()
 
 	err = tmpl.Execute(f, []struct {
-		API  string
-		Kind string
+		API             string
+		Kind            string
+		GracefulSeconds string
```
ymmt2005: int ではだめ？

PR #190, opened 2019-06-14
[ckecli] add sabakan enable subcommand

ymmt2005:
"ckecli sabakan disable" no longer removes sabakan URL.


morimoto-cybozu:
Code modification looks good to me.
Can we check that the integration does not really happen when disabled in mtest or in neco dctest?

ymmt2005:
@morimoto-cybozu We have a test for it in neco/dctest.

PR #228, opened 2019-08-30
Do not recreate resource when patch is not applied

ymmt2005:
CKE recreated resources when applying patch to resources failed.
This turns out too dangerous as CKE may disrupt services.

Patch failure should be handled carefully by humans.

Fixes cybozu-go/neco#501.


zoetrope:
> Patch failure should be handled carefully by humans.

How do I fix the patch failure for dctest?
Should I write codes to delete the resource in dctest?

ymmt2005:
> Should I write codes to delete the resource in dctest?

Yes.

PR #242, opened 2019-09-26
Expose cAdvisor metrics

dulltz:


dulltz:
Close because this PR is no longer needed.

PR #248, opened 2019-10-01
Implement strategies as treating unreachable node.

kfyharukz:
Tp tolerate a small number of problematic nodes in the CKE cluster,
this pull request makes the following changes:

- avoid creating Kubernetes cluster when more than one ssh unconnectable control plane exists.
- skip executing any operations to ssh unconnectable nodes that require ssh connection like starting containers in the nodes.
- avoid using API servers on problematic nodes as endpoints.
- add robustness test that CKE continues to work even if one control plane is down and one other control plane is ssh unconnectable.

### DIFF
--- a/server/node_filter.go
+++ b/server/node_filter.go
@@ -804,14 +804,13 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
 	return false
 }
 
-// SSHNotConnectedNodes returns nodes which are not connected via SSH from targets
-// If targets is nil, it is set as all Nodes in cluster
-func (nf *NodeFilter) SSHNotConnectedNodes(targets []*cke.Node, isCp, isWorker bool) (nodes []*cke.Node) {
+// SSHNotConnectedNodes returns nodes that are not connected via SSH out of targets.
+func (nf *NodeFilter) SSHNotConnectedNodes(targets []*cke.Node, includeControlPlane, includeWorker bool) (nodes []*cke.Node) {
 	for _, n := range targets {
-		if n.ControlPlane && !isCp {
+		if n.ControlPlane && !includeControlPlane {
 			continue
 		}
-		if !n.ControlPlane && !isWorker {
+		if !n.ControlPlane && !includeWorker {
 			continue
 		}
 		if nf.status.NodeStatuses[n.Address].SSHConnected {
@@ -823,14 +822,13 @@ func (nf *NodeFilter) SSHNotConnectedNodes(targets []*cke.Node, isCp, isWorker b
 	return nodes
 }
 
-// SSHConnectedNodes returns nodes which are not connected via SSH from targets
-// If targets is nil, it is set as all Nodes in cluster
-func (nf *NodeFilter) SSHConnectedNodes(targets []*cke.Node, isCp, isWorker bool) (nodes []*cke.Node) {
+// SSHConnectedNodes returns nodes that are connected via SSH out of targets
+func (nf *NodeFilter) SSHConnectedNodes(targets []*cke.Node, includeControlPlane, includeWorker bool) (nodes []*cke.Node) {
 	for _, n := range targets {
-		if n.ControlPlane && !isCp {
+		if n.ControlPlane && !includeControlPlane {
 			continue
 		}
-		if !n.ControlPlane && !isWorker {
+		if !n.ControlPlane && !includeWorker {
 			continue
 		}
 		if !nf.status.NodeStatuses[n.Address].SSHConnected {
--- a/server/node_filter.go
+++ b/server/node_filter.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"fmt"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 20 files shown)


server/node_filter.go
```diff
@@ -783,3 +803,41 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
 
 	return false
 }
+
+// SSHNotConnectedNodes returns nodes which are not connected via SSH from targets
+// If targets is nil, it is set as all Nodes in cluster
```
ymmt2005: > If targets is nil, it is set as all Nodes in cluster

Is this true?

server/node_filter.go
```diff
@@ -783,3 +803,41 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
 
 	return false
 }
+
+// SSHNotConnectedNodes returns nodes which are not connected via SSH from targets
```
ymmt2005: > SSHNotConnectedNodes returns nodes which are not connected via SSH from targets

SSHNotConnectedNodes returns nodes that are not connected via SSH _out of_ targets.

server/node_filter.go
```diff
@@ -783,3 +803,41 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
...
+// If targets is nil, it is set as all Nodes in cluster
+func (nf *NodeFilter) SSHNotConnectedNodes(targets []*cke.Node, isCp, isWorker bool) (nodes []*cke.Node) {
```
ymmt2005: `includeControlPlane`, `includeWorker`.

server/node_filter.go
```diff
@@ -783,3 +803,41 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
...
+			continue
+		}
+
+		nodes = append(nodes, n)
+	}
+	return nodes
+}
+
+// SSHConnectedNodes returns nodes which are not connected via SSH from targets
+// If targets is nil, it is set as all Nodes in cluster
```
ymmt2005: ditto.

server/node_filter.go
```diff
@@ -783,3 +803,41 @@ func nodeIsOutdated(n *cke.Node, current *corev1.Node, taintCP bool) bool {
...
+func (nf *NodeFilter) SSHConnectedNodes(targets []*cke.Node, isCp, isWorker bool) (nodes []*cke.Node) {
```
ymmt2005: ditto.

server/strategy.go
```diff
@@ -20,21 +20,29 @@ func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.ResourceDe
 
 	nf := NewNodeFilter(c, cs)
 
+	// 0. Skip this iteration, if Etcd is not available and one or more CP node(s) are not connected
```
ymmt2005: I think there is no risk to _start_ and _wait_ etcd members even some
nodes are not accessible via SSH.

server/strategy.go
```diff
@@ -20,21 +20,29 @@ func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.ResourceDe
...
 	if ops := riversOps(c, nf); len(ops) > 0 {
 		return ops
 	}
 
-	// 2. Bootstrap etcd cluster, if not yet.
-	if !nf.EtcdBootstrapped() {
-		return []cke.Operator{etcd.BootOp(nf.ControlPlane(), c.Options.Etcd, c.Options.Kubelet.Domain)}
-	}
+	// Etcd boot/start operations run only when all CPs are SSH reachable
+	if len(nf.SSHNotConnectedNodes(nf.cluster.Nodes, true, false)) == 0 {
```
ymmt2005: This `if` should only protect the bootstrapping case.

```go
if !nf.EtcdBootstrapped() {
    if len(nf.SSHNotConnectedNodes(nf.cluster.Nodes, true, false)) == 0 {
        // log etcd bootstrapping cannot go
        return nil
    }
    return []cke.Operator{etcd.BootOp(nf.ControlPlane(), c.Options.Etcd, c.Options.Kubelet.Domain)}
}
```

server/strategy.go
```diff
@@ -168,19 +181,26 @@ func k8sMaintOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.Resource
 
 	ops = append(ops, decideNodeDNSOps(apiServer, c, ks)...)
 
-	cpAddresses := make([]corev1.EndpointAddress, len(nf.ControlPlane()))
-	for i, cp := range nf.ControlPlane() {
-		cpAddresses[i] = corev1.EndpointAddress{
-			IP: cp.Address,
-		}
+	var epAddresses []corev1.EndpointAddress
```
ymmt2005: `cpReadyAddresses`

server/strategy.go
```diff
@@ -168,19 +181,26 @@ func k8sMaintOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.Resource
...
+	for _, n := range nf.HealthyAPIServerNodes() {
+		epAddresses = append(epAddresses, corev1.EndpointAddress{
+			IP: n.Address,
+		})
+	}
+	var epNotReadyAddresses []corev1.EndpointAddress
```
ymmt2005: `cpNotReadyAddresses`

server/strategy.go
```diff
@@ -203,12 +220,18 @@ func k8sMaintOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.Resource
 		ops = append(ops, svcOp)
 	}
 
+	cpReadyAddresses = make([]corev1.EndpointAddress, len(nf.ControlPlane()))
```
ymmt2005: `cpAddresses :=`

PR #253, opened 2019-10-25
Bump version to 1.15.5

binoue:


ymmt2005:
I don't think it is worth a new version.

PR #259, opened 2019-12-06
Add flag volume to prevent start-etcd without initial-cluster

kfyharukz:
When a control plane node of k8s is rebooted, etcd cluster becomes broken.
It happens because the rebooted etcd is started without initial-cluster parameter.

So, in this issue, we add code to create/delete docker volume for flag that etcd-add-member has successfully competed or not.

### DIFF
--- a/op/status.go
+++ b/op/status.go
@@ -58,8 +58,7 @@ func GetNodeStatus(ctx context.Context, inf cke.Infrastructure, node *cke.Node, 
 
 	status.Etcd = cke.EtcdStatus{
 		ServiceStatus: ss[EtcdContainerName],
-		HasData:       etcdVolumeExists,
-		IsAddedMember: isAddedmember,
+		HasData:       etcdVolumeExists && isAddedmember,
 	}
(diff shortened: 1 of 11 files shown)


op/status.go
```diff
@@ -51,9 +51,15 @@ func GetNodeStatus(ctx context.Context, inf cke.Infrastructure, node *cke.Node,
...
 
+	isAddedmember, err := ce.VolumeExists(EtcdAddedMemberVolumeName)
+	if err != nil {
+		return nil, err
+	}
+
 	status.Etcd = cke.EtcdStatus{
 		ServiceStatus: ss[EtcdContainerName],
 		HasData:       etcdVolumeExists,
+		IsAddedMember: isAddedmember,
```
ymmt2005: How about changing `HasData` to `etcdVolumeExists && isAddedmember`
and not adding `IsAddedMember` field?

kfyharukz: It sounds good. I'll refactor it.

PR #278, opened 2019-12-23
Add "ckecli status" and refactor mtest using it.

ymmt2005:
https://github.com/cybozu-go/neco/issues/705

### DIFF
--- a/docs/ckecli.md
+++ b/docs/ckecli.md
@@ -10,18 +10,19 @@ $ ckecli [--config FILE] <subcommand> args...
 | `--config`  | `/etc/cke/config.yml` | config file path    |
 | `--version` |                       | show ckecli version |
 
-`ckecli cluster set FILE`
--------------------------
+## `ckecli cluster`
+
+### `ckecli cluster set FILE`
 
 Set the cluster configuration.
 
-`ckecli cluster get`
---------------------
+### `ckecli cluster get`
 
 Get the cluster configuration.
 
-`ckecli constraints set NAME VALUE`
------------------------------------
+## `ckecli constraints`
+
+### `ckecli constraints set NAME VALUE`
 
 Set a constraint on the cluster configuration.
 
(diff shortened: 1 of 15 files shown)


docs/ckecli.md
```diff
@@ -255,3 +246,14 @@ Set the query variables to search machines in sabakan.
...
+
+## `ckecli status`
+
+Report the internal status of the CKE server.
+
+See [schema.md](schema.md#status)
+
+Example:
+```json
+{"status":"completed","last-updated" : "2019-10-02-00:00:00"}
```
ymmt2005: This is not the serialization format of `time.Time` for JSON.
It should be formatted as RFC3339.

docs/schema.md
```diff
@@ -107,3 +107,13 @@ configuration.
...
+
+<a name="status"></a>
+### `status`
+
+JSON object that has the following fields:
+
+Name           | Type    | Description
+----           | ----    | -----------
+`status`       | string  | CKE server processing status represented as a string. 
+`last-updated` | string  | The Last time when the processing status was updated.
```
ymmt2005: RFC3339 formatted string of the time when the status is recorded.

mtest/run.go
```diff
@@ -430,7 +443,11 @@ func (e checkError) Error() string {
 }
 
 func checkCluster(c *cke.Cluster) error {
-	status, res, err := getClusterStatus(c)
+	/**
```
ymmt2005: Please don't use `/* */` style comment.  It cannot be nested and rarely used.

server/control.go
```diff
@@ -26,6 +26,14 @@ type Controller struct {
 	addon           Integrator
 }
 
+type Status struct {
+	// CKE server processing status represented as a string.
+	Status string `json:"status"`
```
ymmt2005: This can be `ProcessingStatus` instead of `string`.

storage.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/cybozu-go/cke/server"
```
ymmt2005: This should be grouped separately.
If you are using `gopls` in VSCode, this is fixed automatically.

docs/schema.md
```diff
@@ -107,3 +107,13 @@ configuration.
 ### `sabakan/url`
 
 Sabakan URL.
+
+<a name="status"></a>
+### `status`
```
ymmt2005: Wrong title level.

server/strategy.go
```diff
@@ -14,67 +14,85 @@ import (
...
 )
 
-// DecideOps returns the next operations to do.
+// ProcessingStatus represents the processing status of CKE server.
+type ProcessingStatus string
+
+// Processing statuses of CKE server.
+const (
+	StatusUpgrade         ProcessingStatus = "upgrade"
+	StatusRivers                           = "rivers"
```
ymmt2005: The type cannot be omitted.  This just defines a `string` constant.

storage.go
ymmt2005: Actually, this makes a circular reference...

PR #283, opened 2019-12-25
mtest: remove use of loose check.

morimoto-cybozu:
`looseCheck` in the mtest becomes unnecessary, because the completion check of the CKE server's activity is now observable through `ckecli status` without using the version-dependent internal algorithm.
This PR removes it.

### DIFF
--- a/mtest/before.go
+++ b/mtest/before.go
@@ -21,8 +21,6 @@ const (
 func RunBeforeSuite(img string) {
 	if img == "" {
 		img = ckeImageURL
-	} else {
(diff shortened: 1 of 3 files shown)


mtest/before.go
```diff
@@ -21,8 +21,6 @@ const (
 func RunBeforeSuite(img string) {
 	if img == "" {
 		img = ckeImageURL
```
ymmt2005: Please remove this `if` and fix the callers to pass `ckeImageURL`.

PR #286, opened 2020-01-16
Sleep in preStop hook for graceful termination

tapih:
https://github.com/cybozu-go/neco/issues/661


static/cluster-dns.yml
```diff
@@ -94,6 +94,12 @@ spec:
             cpu: 100m
             memory: 70Mi
         args: [ "-conf", "/etc/coredns/Corefile" ]
+        lifecycle:
+          preStop:
+            exec:
```
ymmt2005: I prefer `exec: ["sh", "-c", "sleep", "5"]` shortly.

ymmt2005:
@tapih What about node-dns?

PR #292, opened 2020-02-04
Expose metrics to prometheus

tapih:
https://github.com/cybozu-go/neco/issues/747


docs/metrics.md
```diff
@@ -0,0 +1,12 @@
+Metrics
+=======
+
+CKE exposes the following metrics with the Prometheus format. The listen address can be configured by the CLI flag (see [here](cke.md#Usage)). All these metrics are prefixed with `cke_`
+
+| Name              | Description                                      | Type  | Labels                             |
+| ----------------- | ------------------------------------------------ | ----- | ---------------------------------- |
+| operation_running | True(=1) if any operations are running.          | Gauge |                                    |
+| boot_leader       | True(=1) if the boot server is the leader of CKE | Gauge |                                    |
```
ymmt2005: `boot_` looks extraneous.

docs/metrics.md
```diff
@@ -0,0 +1,12 @@
...
+| node_info         | The Control Plane and Worker info                | Gauge | address, rack, role, control_plane |
```
ymmt2005: This is completely not a metric.

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+	storage *Storage
+}
+
+// NewCollector returns a new Collector.
+func NewCollector(client *v3.Client) *Collector {
+	return &Collector{
+		metrics: map[string]Metric{
+			"operation_phase": {
+				collectors:  []prometheus.Collector{operationPhase, operationPhaseTimestampSeconds},
+				isAvailable: alwaysAvailable,
```
ymmt2005: Really?  I think this should only be available when the instance is the leader.

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+type logger struct{}
+
+func (l logger) Println(v ...interface{}) {
+	log.Error(fmt.Sprint(v...), nil)
+}
+
+const (
+	namespace     = "cke"
```
ymmt2005: The name `namespace` is too generic for the package `cke`.
Consider creating `metrics` package and move this file under it.
Splitting this file should also be considered as this file is a bit large.

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+	scrapeTimeout = time.Second * 10
```
ymmt2005: The default scrape timeout value of Prometheus is also 10 seconds.
Considering that, this value should be shortened, shouldn't this?

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+)
+
+// Metric represents collector and availability of metric.
+type Metric struct {
```
ymmt2005: I think `MetricGroup` is more appropriate.

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+	collectors  []prometheus.Collector
+	isAvailable func(context.Context, *Storage) (bool, error)
+}
+
+// Collector is a metrics collector for CKE.
+type Collector struct {
```
ymmt2005: Why should this be exported?

metrics.go
```diff
@@ -0,0 +1,217 @@
...
+	metrics map[string]Metric
+	storage *Storage
+}
+
+// NewCollector returns a new Collector.
+func NewCollector(client *v3.Client) *Collector {
```
ymmt2005: Can `*Collector` be replaced with `prometheus.Collector`?

pkg/cke/main.go
```diff
@@ -98,9 +99,25 @@ func main() {
...
+	mux.Handle("/metrics", metricsHandler)
+	ms := &well.HTTPServer{
+		Server: &http.Server{
+			Addr:    *flgMetrics,
+			Handler: mux,
+		},
+	}
+	ms.ListenAndServe()
+
+	// API server
```
ymmt2005: Why don't you merge the two?

storage.go
```diff
@@ -490,7 +501,7 @@ func (s Storage) GetLeaderHostname(ctx context.Context) (string, error) {
 	}
 
 	if len(resp.Kvs) == 0 {
-		return "", errors.New("no leader")
```
ymmt2005: Why do you change this line?
`ErrLeaderNotExist` is used only here.  It seems not worthy of a constant.

metrics.go
morimoto-cybozu: That's right.  I'm updating the code and adding the following note to the doc.

> All metrics but `leader` are available only when the server is the leader of CKE.

metrics.go
morimoto-cybozu: Absolutely, it should be shortened.
The purpose I applied this timeout was to reclaim resources as soon as possible *after* the scraping timeout.  Now I'm reminded that the collector should return as many metrics as possible even when some collection goroutine is delayed.  The shorter timeout will enable partial responses.

metrics.go
morimoto-cybozu: There is no need to export.  Fixed.

storage.go
morimoto-cybozu: `ErrLeaderNotExist` was once used in multiple code portions, which have been removed now.  Reverted.

pkg/cke/main.go
morimoto-cybozu: I supposed that the REST API of CKE was too complicated to be multiplexed with `/metrics`.  Actually CKE has just two endpoints.  Merged.

docs/metrics.md
```diff
@@ -0,0 +1,19 @@
+Metrics
+=======
+
+CKE exposes the following metrics with the Prometheus format. The listen address can be configured by the CLI flag (see [here](cke.md#Usage)). The path is fixed to `/metrics`. All these metrics are prefixed with `cke_`
```
ymmt2005:
```suggestion
CKE exposes the following metrics with the Prometheus format at `/metrics` REST API endpoint.  All these metrics are prefixed with `cke_`
```

ymmt2005:
@morimoto-cybozu Could you please resolve the conflict?

morimoto-cybozu:
@ymmt2005 Yeah, sure!

PR #299, opened 2020-03-05
k8s 1.17: Server Side Apply

dulltz:
Upgrade K8s to 1.17

##### Commons:
- Update container images
- Update tools for mtest
- Update tools for conformance test
- Upgrade library for k8s 1.17.3

##### Changed:
- Use server-side apply for resource-apply-op
    - If the target resources have never been server-side applied, CKE apply forcibly.


dulltz:
@ymmt2005 
> I'd think "true" is clearer.

We cannot  hard-code as `"true"`,
because we use this function in both v3-upgrade-op and resource-apply-op.
In the latter case, we should not apply resources forcibly.

ymmt2005:
@dulltz 
You in fact hard-coded `"true"` by using `strconv`.
```go
strconv.FormatBool(params.forceConflicts)
```

dulltz:
@ymmt2005 
> You in fact hard-coded "true" by using strconv.
> ```
> strconv.FormatBool(params.forceConflicts)
> ```

On, really? In my understanding, I intended to put two kinds of bool values.

We call `ApplyResource()` at two lines:
-   [In upgradeToVersion3Command](https://github.com/cybozu-go/cke/pull/299/files#diff-46d9726d904478f03dc2846c8456261fR53)
-   [In resourceApplyOp](https://github.com/cybozu-go/cke/pull/299/files#diff-a8cdf5fe7131e655647fbff97a663e19R46)
    
In the former case we substitute `true` for `forceConflicts` argument,
and in the latter case we substitute `false` for it.

Then in `ApplyResource()`,  we set `forceConflicts` argument in `applyParams.forceConflicts`.

ymmt2005:
Ah, I see.  sorry for bothering.

op/upgrade_v2.go
```diff
@@ -7,31 +7,29 @@ import (
...
-		u.current = "2"
-		return UpgradeToVersion2Command(u.nodes)
-	default:
+func (u *v2UpgradeOp) NextCommand() cke.Commander {
+	if u.completed {
 		return nil
 	}
+	u.completed = true
+	return UpgradeToVersion2Command(u.nodes)
 }
```
ymmt2005: I think `upgradeOp` should be kept intact.

Just adding `case "2"` to `NextCommand()` is enough, isn't it?

op/upgrade_v3.go
```diff
@@ -0,0 +1,70 @@
+package op
+
+import (
+	"context"
+	"strings"
+
+	"github.com/cybozu-go/cke"
+)
+
+type v3UpgradeOp struct {
```
ymmt2005: I don't think `v3UpgradeOp` is needed.

The original `upgradeOp` can determine what to do for the current cluster version.

pkg/apply_gen/main.go
```diff
@@ -21,60 +21,43 @@ import (
...
-	current, err := getFunc(o.Name, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		_, err = createFunc(o)
-		return err
-	}
-	if err != nil {
-		return err
-	}
-
+func apply{{ .Kind }}(o *{{ .API }}.{{ .Kind }}, rev int64, client rest.Interface, params *applyParams) error {
```
ymmt2005: The motivation when `apply` was first implemented as template-rendering was
to use typed REST client for safer operations.  Now that `rest.Interface`, a non-typed
REST client, is used, there is no reasons to use templates.

Would you please implement this just as a pure Go function?

server/strategy.go
```diff
@@ -19,9 +19,19 @@ import (
 func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, resources []cke.ResourceDefinition) ([]cke.Operator, cke.OperationPhase) {
 	nf := NewNodeFilter(c, cs)
 
-	// 0. Execute upgrade opeartion if necessary
+	// 0. Execute upgrade operation if necessary
 	if cs.ConfigVersion != cke.ConfigVersion {
-		return []cke.Operator{op.UpgradeOp(cs.ConfigVersion, nf.ControlPlane())}, cke.PhaseUpgrade
+		switch cs.ConfigVersion {
```
ymmt2005: can be compacted.  though this should be reverted as commented above.

storage.go
```diff
@@ -77,11 +77,11 @@ func (s Storage) GetConfigVersion(ctx context.Context) (string, error) {
 }
 
 // PutConfigVersion sets the current configuration version of the Kubernetes cluster.
-func (s Storage) PutConfigVersion(ctx context.Context, leaderKey string) error {
+func (s Storage) PutConfigVersion(ctx context.Context, leaderKey, version string) error {
 	resp, err := s.Txn(ctx).
 		If(clientv3util.KeyExists(leaderKey)).
 		Then(
-			clientv3.OpPut(KeyConfigVersion, ConfigVersion),
+			clientv3.OpPut(KeyConfigVersion, version),
```
ymmt2005: This should be reverted.

pkg/apply_gen/main.go
dulltz: Agree with you, we will refactor this code in the next PR of this task.

status.go
```diff
@@ -74,8 +82,8 @@ func (s KubernetesClusterStatus) IsReady(cluster *Cluster) bool {
 }
 
 // SetResourceStatus sets status of the resource.
-func (s KubernetesClusterStatus) SetResourceStatus(rkey string, annotations map[string]string) {
-	s.ResourceStatuses[rkey] = annotations
+func (s KubernetesClusterStatus) SetResourceStatus(rkey string, rs ResourceStatus) {
+	s.ResourceStatuses[rkey] = rs
```
ymmt2005: This function may become more convenient if it takes arguments like this:

```go
func (s KubernetesClusterStatus) SetResourceStatus(rkey string, ann map[string]string, isManaged bool) {
    s.ResourceStatuses[rkey] = ResourceStatus{Annotations: ann, HasBeenSSA: isManaged}
}
```

masa213f:
@zoetrope 
As an example, it is better to use officially released images and binaries, I think.
So, in this PR, I don't want to fix it (or update to use `quay.io/cybozu/cke:1.16`).

We will change the tag at officially CKE release.

PR #301, opened 2020-03-10
k8s 1.17: Import types from official repos

llamerada-jp:
- Import types from k/apiserver k/kube-scheduelr, k/kubelet.
- Stop using `rbac.authorization.k8s.io/v1beta1` because this version has been obsoleted in the range of CKE support versions 1.15-1.17.


op/k8s/config.go
```diff
@@ -41,128 +44,43 @@ func kubeletKubeconfig(cluster string, n *cke.Node, caPath, certPath, keyPath st
...
 
-// MarshalJSON implements json.Marshaler.
-func (a KubeletAuthentication) MarshalJSON() ([]byte, error) {
-	v, err := a.MarshalYAML()
-	if err != nil {
-		return nil, err
+func newEncryptionConfiguration() apiserverv1.EncryptionConfiguration {
+	return apiserverv1.EncryptionConfiguration{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apiserver.config.k8s.io/v1",
```
ymmt2005: Use predefined constants like this:
https://github.com/kubernetes/apiserver/blob/master/pkg/apis/config/v1/register.go#L25

ymmt2005: Or, it'd be better to use `scheme` and register the types to the scheme,
then encode the structs into JSON using `scheme`.

masa213f:
@ymmt2005 
Which file remains? I think I had deleted all.

ymmt2005:
@masa213f 
I thought `scheduler/k8s_types.go` was not deleted, but it was in fact deleted.
GitHub shows as if the file is just renamed... Sorry.

PR #309, opened 2020-04-06
Revert the Makefile target to run sonobuoy with Vagrant

tapih:
https://github.com/cybozu-go/neco/issues/833

### DIFF
--- a/sonobuoy/Makefile
+++ b/sonobuoy/Makefile
@@ -54,7 +54,20 @@ run: bin/cke bin/ckecli bin/kubectl
 	./bin/kubectl apply -f ../example/calico.yaml
 	for i in $$(seq 300); do sleep 1; if [ "$$(./bin/kubectl get nodes --no-headers | grep Ready | wc -l)" = 3 ]; then break; fi; done
 	for i in $$(seq 300); do sleep 1; if [ "$$(./bin/kubectl get -n kube-system deployments/cluster-dns -o json | jq .status.readyReplicas)" = 2 ]; then break; fi; done
-
+run-on-vagrant: bin/cke bin/ckecli bin/kubectl
+	mkdir -p etcd-data
+	/usr/local/bin/docker-compose up -d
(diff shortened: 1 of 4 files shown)


sonobuoy/Makefile
```diff
@@ -54,7 +56,6 @@ run: bin/cke bin/ckecli bin/kubectl
 	./bin/kubectl apply -f ../example/calico.yaml
 	for i in $$(seq 300); do sleep 1; if [ "$$(./bin/kubectl get nodes --no-headers | grep Ready | wc -l)" = 3 ]; then break; fi; done
 	for i in $$(seq 300); do sleep 1; if [ "$$(./bin/kubectl get -n kube-system deployments/cluster-dns -o json | jq .status.readyReplicas)" = 2 ]; then break; fi; done
-
```
ymmt2005, line 57: Keep the blank line to separate targets.

sonobuoy/Makefile
```diff
@@ -63,11 +64,31 @@ sonobuoy: bin/sonobuoy
 	./bin/sonobuoy delete
 
 clean:
+	-vagrant destroy -f
```
ymmt2005, line 67: I don't think adding `vagrant destroy` to `clean` is the right thing.
Moreover, it should _not_ ignore the errors from `vagrant` if this were the right place.

Instead, I'd suggest that users should run `vagrant destroy` manually after
running `make run-on-vagrant`.

sonobuoy/README.md
```diff
@@ -1,7 +1,7 @@
 Conformance test for CKE
 ========================
 
-Follow the instructions to run [Sonobuoy][] for conformance test of certified Kubernetes.
+Follow the instructions to run [Sonobuoy][] for conformance test of certified Kubernetes with Vagrant.
```
ymmt2005, line 4: Since there is `make sonobuoy` to run Sonobuoy, this line looks odd.

PR #312, opened 2020-04-09
Use dynamic client for SSA

### LINKED ISSUE #311: Failed to update squid deployment with SSA
**Describe the bug**

After applying [this release](https://github.com/cybozu-go/neco/releases/tag/release-2020.04.08-8660) to the staging environment, CKE failed to apply the squid deployment.

This resource is applied with SSA and the failure is caused by the SSA conflict.
The corresponding CKE operation is `resource-apply` and the error message is `Apply failed with 1 conflict: conflict with \"before-first-apply\" using apps/v1: .spec.template.spec.containers[name=\"squid\"].image`.

**Environments**
- Version: v1.17.1

**To Reproduce**
1. Run CKE v1.17.0
2. Update CKE to v1.17.1

**Expected behavior**
`squid` is deployed with no error.

**Additional context**

ymmt2005:
For Server-Side-Apply (SSA), resources must not be decoded
into statically typed Go structs such as `appsv1.DaemonSet`
because doing so will add unintended fields to JSON when
the struct is encoded back into JSON for SSA.

Instead, they should be decoded into `unstructured.Unstructured`
and will be encoded as is with the dynamic client.

Fixes #311


op/resource.go
```diff
@@ -42,11 +46,21 @@ func (o *resourceApplyOp) Targets() []string {
...
+	if err != nil {
+		return err
+	}
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(dc))
+
+	dyn, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return err
+	}
+	return cke.ApplyResource(dyn, mapper, o.resource.Definition, o.resource.Revision, true)
```
ymmt2005, line 63: Passing `true` to force server-side apply due to kubernetes/kubernetes#89954

PR #325, opened 2020-05-20
sabakan: append used/unused machines info to Generator. 

masa213f:
This PR is a preparation to make sabakan integration's strategy better.

Currently, the intermediate state to decide the next `cke.Cluster` is separated into `Generator.unusedMachines` and `updateOp.{cps, workers}`. So, a reference to them becomes complex and difficult to understand.

With this PR, the intermediate state becomes aggregated into struct `Generator` only.


sabakan/generator.go
```diff
@@ -503,21 +508,12 @@ func (g *Generator) increaseControlPlane() (*updateOp, error) {
...
-		workers[i] = g.machineMap[n.Address]
-	}
-
 	op := &updateOp{
-		name:    "increase control plane",
-		cps:     cps,
-		workers: workers,
+		name: "increase control plane",
 	}
+	g.nextControlPlanes = g.nodesToMachines(true)
```
kfyharukz, line 514: func name  nodesToMachines is slightly ambiguous. Adding the word "current" is better.

sabakan/generator.go
```diff
@@ -526,36 +522,27 @@ func (g *Generator) decreaseControlPlane() (*updateOp, error) {
...
+		m = g.deselectControlPlane(numCPs)
 
-		if g.constraints.MaximumWorkers == 0 || len(op.workers) < g.constraints.MaximumWorkers {
+		if g.constraints.MaximumWorkers == 0 || len(g.nextWorkers) < g.constraints.MaximumWorkers {
 			op.demoteControlPlane(m)
+			g.nextWorkers = append(g.nextWorkers, m)
 			continue
 		}
 
+		g.nextUnused = append(g.nextUnused, m)
```
kfyharukz, line 542: How should CKE work when the worker num exceeds MaximumWorkers due to demoting cp to a worker?
Should CKE remain the worker in the cluster?

ymmt2005: Doing so would remove a running worker node to keep satisfying `MaximumWorkers`.
It is not clear which is better.

We may improve the integration strategy not to reach `MaximumWorkers` to keep
a room for demoting control planes.

PR #340, opened 2020-05-29
Stop all apiserver at the same time when restarting.

kfyharukz:
This PR is to prevent that different version `kube-apiserver`s exist at the same time by stop `kube-apiserver` simultaneously.
ref: https://github.com/cybozu-go/neco/issues/880


op/common/container.go
```diff
@@ -109,42 +109,52 @@ func (c runContainerCommand) Command() cke.Command {
...
-		return nil
-	}
-	// Inspect returns ServiceStatus for the named container.
-	statuses, err := ce.Inspect([]string{c.name})
-	if err != nil {
-		return err
-	}
-	if st, ok := statuses[c.name]; ok && st.Running {
-		err = ce.Stop(c.name)
+	for _, n := range c.nodes {
```
ymmt2005, line 127: Make this parallel, just like `Run`.

ymmt2005:
@kfyharukz @dulltz 
You need to modify these lines too:
https://github.com/cybozu-go/cke/blob/master/op/k8s/apiserver_restart.go#L53-L55

ymmt2005:
@kfyharukz @dulltz 
One more point.  It is better to exchange `APIServerBootOp` and `APIServerRestartOp`
in [server/stragegy.go](https://github.com/cybozu-go/cke/blob/master/server/strategy.go#L102-L107) because otherwise there could be mixed API server versions
if CKE is killed during `APIServerRestartOp` and only part of API servers are stopped.

op/common/container.go
dulltz: OK!

PR #343, opened 2020-05-30
Fix race in server.Controller.GetClusterStatus

ymmt2005:
A concurrent reference of a map caused the race because
other goroutines are mutating the map at the same time.
https://app.circleci.com/pipelines/github/cybozu-go/cke/1085/workflows/92fc41ee-55b3-4877-8098-c5edf86ddcd4/jobs/11128/steps

In fact, the map does not need to be updated, so fixing it that way.

In addition, remove redundant tests from `mtest/operators.go`.

### DIFF
--- a/mtest/operators.go
+++ b/mtest/operators.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/cybozu-go/cke"
-	"github.com/cybozu-go/cke/op"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


mtest/operators.go
```diff
@@ -332,66 +331,6 @@ func TestOperators(isDegraded bool) {
...
-
-		By("Adding a scheduler extender")
-		// this will run these ops:
-		// - SchedulerRestartOp
-		cluster.Options.Scheduler.Extenders = []string{"urlPrefix: http://127.0.0.1:8000"}
-		clusterSetAndWait(cluster)
-
-		stdout, stderr, err := execAt(node1, "jq", "-r", "'.extenders[0].urlPrefix'", op.PolicyConfigPath)
-		Expect(err).NotTo(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
-		Expect(strings.TrimSpace(string(stdout))).To(Equal("http://127.0.0.1:8000"))
```
llamerada-jp, line 344: Didn't confirmed values in PolicyConfigPath at strategy test.go
Don't you need to check the value of the changed result for each case?
https://github.com/cybozu-go/cke/blob/master/server/strategy_test.go#L745-L757

ymmt2005: If you want to check whether the value is actually retrieved by GetClusterStatus,
write the test for GetClusterStatus in a separate file.  Tests in `operators.go`
are meant to test operations such as `SchedulerRestartOp`, and `SchedulerRestartOp`
was tested unnecessarily too many times.

PR #346, opened 2020-06-03
Avoid node duplication due to regeneration after update

umezawatakeshi:
Avoid node duplication due to `Generator.Regenere` after `Generator.Update`.
And add validation for node duplication.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -191,8 +191,7 @@ func (c *Cluster) Validate(isTmpl bool) error {
 
 	nodeAddressSet := make(map[string]struct{})
 	for _, n := range c.Nodes {
-		_, ok := nodeAddressSet[n.Address]
-		if ok {
+		if _, ok := nodeAddressSet[n.Address]; ok {
 			return errors.New("duplicate node address: " + n.Address)
 		}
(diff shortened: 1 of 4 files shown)


cluster.go
```diff
@@ -189,6 +189,15 @@ func (c *Cluster) Validate(isTmpl bool) error {
 		}
 	}
 
+	nodeAddressSet := make(map[string]int)
```
ymmt2005, line 192: For the best performance, it is advised to use `map[string]struct{}`.

For the best convenience, it is good to use `map[string]bool` because it can be used as:

```go
    if nodeAddressSet[n.Address] {
        return err
    }
```

cluster.go
```diff
@@ -189,6 +189,15 @@ func (c *Cluster) Validate(isTmpl bool) error {
 		}
 	}
 
+	nodeAddressSet := make(map[string]struct{})
+	for _, n := range c.Nodes {
+		_, ok := nodeAddressSet[n.Address]
+		if ok {
```
ymmt2005, line 195: This can be simplified:
```suggestion
		if _, ok := nodeAddressSet[n.Address]; ok {
```

PR #348, opened 2020-06-03
Fix template validation bug introduced in #346

ymmt2005:


cluster.go
```diff
@@ -223,12 +222,17 @@ func (c *Cluster) Validate(isTmpl bool) error {
 	return nil
 }
 
-func (c *Cluster) validateNode(n *Node, isTmpl bool, fldPath *field.Path) error {
-	if !isTmpl {
+func validateNode(n *Node, isTmpl bool, fldPath *field.Path) error {
+	if isTmpl {
+		if len(n.Address) != 0 {
+			return errors.New("address is not empty: " + n.Address)
+		}
```
ymmt2005, line 229: why do you check this?
anyways, the address will be overwritten.

ymmt2005:
@masa213f 
Please add `if err := c.Validate(false); err != nil ...` to this line
https://github.com/cybozu-go/cke/blob/master/sabakan/generator.go#L316

This makes the template instantiation super safe.

cluster.go
masa213f: To check operation errors.

If the operator mistakenly set the CKE cluster definition (the result of `ckecli cluster get`) as a sabakan template, this check will cause an error.

masa213f:
@ymmt2005 
Thanks. The comments are reflected.

ymmt2005:
LGTM

PR #373, opened 2020-10-12
node-dns: change terminationGracePeriodSeconds from 0 to 1

ymmt2005:
With the value of 0, Kubernetes won't try to kill the process.


morimoto-cybozu (approved):
paired

PR #385, opened 2020-11-19
reboot: add reboot_queue_entries metrics

yokaze:
This PR adds `reboot_queue_entries` metrics to monitor the progress of reboot operation.

### DIFF
--- a/server/control.go
+++ b/server/control.go
@@ -266,15 +266,15 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
 		return err
 	}
 
-	reboot, err := inf.Storage().GetRebootsFrontEntry(ctx)
-	switch err {
-	case nil:
-	case cke.ErrNotFound:
-		reboot = nil
-	default:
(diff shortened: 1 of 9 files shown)


server/control.go
```diff
@@ -266,15 +266,15 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
...
-	case nil:
-	case cke.ErrNotFound:
-		reboot = nil
-	default:
+	re, err := inf.Storage().GetRebootsEntries(ctx)
+	if err != nil {
 		return err
 	}
-
+	metrics.UpdateReboot(len(re))
```
ymmt2005, line 273: Why don't you update this metric inside Collect?

ymmt2005 (changes requested):

server/control.go
ymmt2005, line 273: IIUC, this metric should also be updated every time the server removes an entry from the queue.

ymmt2005 (changes requested):

ymmt2005 (approved):
LGTM

PR #388, opened 2020-11-27
Remove defaulting for kubelet OOMScoreAdj.

kfyharukz:
Currently, `kubelet` is executed with `oom-score-adj=-1000` which means the OOM killer doesn't kill the `kubelet` if it makes a memory leak.
This PR removes the option `oom-score-adj`, which results to use the default value `-999`, then the OOM killer can kill the `kubelet`.


ymmt2005 (approved):
LGTM

PR #396, opened 2020-12-03
reboot: enable/disable reboot queue processing

yokaze:
This PR adds `ckecli rq enable` and `ckecli rq disable` commands.
They allow us to suspend/resume reboot queue processing.

### DIFF
--- a/docs/ckecli.md
+++ b/docs/ckecli.md
@@ -43,6 +43,7 @@ $ ckecli [--config FILE] <subcommand> args...
 - [`ckecli ssh [user@]NODE [COMMAND...]`](#ckecli-ssh-usernode-command)
 - [`ckecli scp [-r] [[user@]NODE1:]FILE1 ... [[user@]NODE2:]FILE2`](#ckecli-scp--r-usernode1file1--usernode2file2)
 - [`ckecli reboot-queue`, `ckecli rq`](#ckecli-reboot-queue-ckecli-rq)
+  - [`ckecli reboot-queue enable|disable`](#ckecli-reboot-queue-enabledisable)
   - [`ckecli reboot-queue add FILE`](#ckecli-reboot-queue-add-file)
   - [`ckecli reboot-queue list`](#ckecli-reboot-queue-list)
   - [`ckecli reboot-queue cancel INDEX`](#ckecli-reboot-queue-cancel-index)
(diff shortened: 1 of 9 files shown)


docs/ckecli.md
```diff
@@ -277,6 +278,10 @@ Copy files between hosts via scp.
 
 `rq` is an alias of `reboot-queue`.
 
+### `ckecli reboot-queue enable|disable`
+
+Enable/Disables processing reboot queue entries.
```
ymmt2005, line 283:
```suggestion
Enable/Disable processing reboot queue entries.
```

storage.go
```diff
@@ -687,6 +688,30 @@ func (s Storage) GetSabakanURL(ctx context.Context) (string, error) {
...
+		return false, err
+	}
+	if resp.Count == 0 {
+		return false, nil
+	}
+
+	if bytes.Equal([]byte("true"), resp.Kvs[0].Value) {
+		return true, nil
+	}
+	return false, nil
```
ymmt2005, line 704:
```suggestion
	return bytes.Equal([]byte("true"), resp.Kvs[0].Value), nil
```

storage.go
```diff
@@ -687,6 +688,30 @@ func (s Storage) GetSabakanURL(ctx context.Context) (string, error) {
...
+}
+
+// EnableRebootQueue enables reboot queue processing when flag is true.
+// When flag is false, reboot queue is not processed.
+func (s Storage) EnableRebootQueue(ctx context.Context, flag bool) error {
+	val := fmt.Sprint(!flag)
```
ymmt2005, line 710: Please be more explicit.
```go
var val string
if flag {
    val = "true"
} else {
    val = "false"
}
```

ymmt2005 (changes requested):

server/control.go
```diff
@@ -266,14 +266,20 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
...
-	metrics.UpdateReboot(len(re))
-	var reboot *cke.RebootQueueEntry
-	if len(re) > 0 {
-		reboot = re[0]
+	if !disabled {
+		re, err := inf.Storage().GetRebootsEntries(ctx)
+		if err != nil {
+			return err
+		}
+		metrics.UpdateReboot(len(re))
```
ymmt2005, line 279: When disabled, the metrics will not be updated.  Is this good behavior?

yokaze: sorry, fixed

ymmt2005 (approved):
LGTM

yokaze:
It should be merged with squash

PR #402, opened 2021-01-08
wip: add kubelet configuration validation

binoue:


morimoto-cybozu:
As of CKE 1.19, the old style configuration is no longer supported.
The new style configuration is automatically validated by typed unmarshal, so this task can be cancelled.

PR #429, opened 2021-03-11
server: add AuditLogPath

yokaze:
This PR adds `audit_log_path` to `APIServerParams`.
Audit logs can be sent to either of journal log or file.
- `audit_log_path` is empty -> journal log
- `audit_log_path` is set -> specified file

`--audit-log-maxage`, `--audit-log-maxbackup`, `--audit-log-maxsize` can be set using `extra_args`.

### DIFF
--- a/op/k8s/apiserver_restart.go
+++ b/op/k8s/apiserver_restart.go
@@ -84,7 +84,7 @@ func (o *apiServerRestartOp) NextCommand() cke.Commander {
 		}
 		paramsMap := make(map[string]cke.ServiceParams)
 		for _, n := range o.nodes {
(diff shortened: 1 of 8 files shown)


op/k8s/apiserver_restart.go
```diff
@@ -311,6 +315,13 @@ func APIServerParams(controlPlanes []*cke.Node, advertiseAddress, serviceSubnet
 				Propagation: "",
 				Label:       cke.LabelShared,
 			},
+			{
+				Source:      "/var/log/audit",
```
ymmt2005, line 319: This should not be embedded in CKE.
Instead, users should mount a host directory by `extra_binds`.

ymmt2005 (changes requested):

ymmt2005 (approved):
LGTM

PR #435, opened 2021-03-22
Fix release-image workflow

binoue:
release-image workflow failed
https://github.com/cybozu-go/cke/runs/2162971053?check_suite_focus=true

This failure was caused by testing without running etcd container.
However, as the test is redundant, this PR deletes the test.

Signed-off-by: binoue <banji-inoue@cybozu.co.jp>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,16 +9,13 @@ all: test
 setup:
 	curl -fsL https://github.com/etcd-io/etcd/releases/download/v$(ETCD_VERSION)/etcd-v$(ETCD_VERSION)-linux-amd64.tar.gz | sudo tar -xzf - --strip-components=1 -C /usr/local/bin etcd-v$(ETCD_VERSION)-linux-amd64/etcd etcd-v$(ETCD_VERSION)-linux-amd64/etcdctl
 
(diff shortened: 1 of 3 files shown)


ymmt2005 (changes requested):
I don't think the test itself is useful because the tagged main branch should have passed the tests.

Makefile
```diff
@@ -9,23 +9,29 @@ all: test
 setup:
 	curl -fsL https://github.com/etcd-io/etcd/releases/download/v$(ETCD_VERSION)/etcd-v$(ETCD_VERSION)-linux-amd64.tar.gz | sudo tar -xzf - --strip-components=1 -C /usr/local/bin etcd-v$(ETCD_VERSION)-linux-amd64/etcd etcd-v$(ETCD_VERSION)-linux-amd64/etcdctl
 
-.PHONY: test
-test: test-tools
+.PHONY: lint
```
ymmt2005, line 12: This isn't a lint thing, actually.  Let's merge this with `test` target.

ymmt2005 (approved):
LGTM

PR #437, opened 2021-03-25
fix image name

binoue:
Signed-off-by: binoue <banji-inoue@cybozu.co.jp>

### DIFF
--- a/.github/workflows/release-images.yaml
+++ b/.github/workflows/release-images.yaml
@@ -68,7 +68,7 @@ jobs:
     name: Release on GitHub
     runs-on: ubuntu-20.04
     container:
-      image: 1.16.2.1_focal
+      image: golang:1.16.2.1_focal
     needs: release-cke-image
     steps:
       - name: Download sonobuoy test
--- a/.github/workflows/release-images.yaml
+++ b/.github/workflows/release-images.yaml
@@ -68,7 +68,7 @@ jobs:
     name: Release on GitHub
     runs-on: ubuntu-20.04
     container:
-      image: golang:1.16.2.1_focal
+      image: quay.io/cybozu/golang:1.16-focal
     needs: release-cke-image
     steps:
       - name: Download sonobuoy test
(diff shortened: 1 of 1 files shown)


ymmt2005 (approved):

.github/workflows/release-images.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
     name: Release on GitHub
     runs-on: ubuntu-20.04
     container:
-      image: 1.16.2.1_focal
+      image: golang:1.16.2.1_focal
```
ymmt2005, line 71: Could you tell me why you didn't use `actions/setup-go`?

ymmt2005, line 71: Did you mean?
```suggestion
      image: quay.io/cybozu/golang:1.16-focal
```

binoue: This release workflow is using `ghr` to GitHub release because [the official action](https://github.com/actions/create-release) to release was archived.

`quay.io/cybozu/golang:1.16-focal` container already contains `ghr`, so I thought releasing would be easy if we use the image as the job's base image. 
(`golang:1.16.2.1_focal` is a mistake)

binoue: Thank you for pointing out. Yes that is a mistake.
However, by checking this job in my repo, specifying the quay image in the field will not work because the account is not logged in as a quay user. 

```
Checking docker version
Clean up resources from previous jobs
Create local container network
Starting job container
  /usr/bin/docker pull quay.io/cybozu-go/golang:1.16-focal
  Error response from daemon: unauthorized: access to the requested resource is not authorized
  Warning: Docker pull failed with exit code 1, back off 5.233 seconds before retry.
  /usr/bin/docker pull quay.io/cybozu-go/golang:1.16-focal
  Error response from daemon: unauthorized: access to the requested resource is not authorized
  Warning: Docker pull failed with exit code 1, back off 2.296 seconds before retry.
  /usr/bin/docker pull quay.io/cybozu-go/golang:1.16-focal
  Error response from daemon: unauthorized: access to the requested resource is not authorized
  Error: Docker pull failed with exit code 1
```

The reason why we are trying to use the image is just that that already contains `ghr`.
But as using the image as a base is impossible, it is required to download and use it on the VM.

binoue: Thank you. Yes that must be `quay.io/cybozu/golang:1.16-focal` and this would work.

yokaze (approved):
LGTM

Issue #432, opened 2021-03-16
vault token was revoked


zoetrope:
**Describe the bug**

There might be a problem that the vault token is revoked if a CKE instance is down for a long time.

`ckecli history` showed the following logs:
```
{
    "id": "30906",
    "status": "cancelled",
    "operation": "etcd-wait-cluster",
    "command": {
        "name": "wait-etcd-sync",
        "target": "https://10.69.1.136:2379,https://10.69.0.14:2379,https://10.69.0.10:2379"
    },
    "targets": [
        "https://10.69.1.136:2379",
        "https://10.69.0.14:2379",
        "https://10.69.0.10:2379"
    ],
    "info": "",
    "error": "Error making API request.\n\nURL: GET https://127.0.0.1:8200/v1/cke/ca-etcd-client/roles/admin\nCode: 403. Errors:\n\n* permission denied",
    "start-at": "2021-03-16T05:46:53.049466554Z",
    "end-at": "2021-03-16T05:47:11.072880945Z"
}
```

**Environments**
- Version:
- OS:

**To Reproduce**

**Expected behavior**

If the vault returns 403, CKE should exit.

**Additional context**

PR #445, opened 2021-03-30
Exit CKE when vault returns 403

masa213f:
https://github.com/cybozu-go/cke/issues/432

### DIFF
--- a/server/control.go
+++ b/server/control.go
@@ -229,7 +229,7 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
 	if err != nil {
 		// When the vault token is revoked, the following error will be returned. In this case, CKE can not continue any operations.
 		// Error: "Error making API request.\n\nURL: GET <<URL>>\nCode: 403. Errors:\n\n* permission denied"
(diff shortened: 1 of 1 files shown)


server/control.go
```diff
@@ -227,6 +227,15 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
 
 	inf, err := cke.NewInfrastructure(ctx, cluster, storage)
 	if err != nil {
+		// When the vault token is revoked, the following error will be returned. In this case, CKE can not continue any operations.
+		// Error: "Error making API request.\n\nURL: GET <<URL>>\nCode: 403. Errors:\n\n* permission denied"
+		if strings.Contains(err.Error(), "Code: 403.") {
```
binoue, line 232: I think checking "Code: 403.” in the error string has a risk of being outdated.
Why don't you check only "403"?

masa213f: That's good. I will change it.

It's a trade-off for false positives.
But even if CKE terminates due to a false positive, it won't cause any problems.

yokaze (approved):

Issue #441, opened 2021-03-26
CKE hangs up when etcd session is closed


masa213f:
**Describe the bug**

When etcd stops, CKE hangs because closing of `concurrency.Session` is not checked.
This causes a failure of the leader election.

**Environments**
- Version:
- OS:

**To Reproduce**

**Expected behavior**

**Additional context**

PR #446, opened 2021-03-31
Watch session and leader key

masa213f:
see https://github.com/cybozu-go/cke/issues/441

When the etcd process which connected from clients stops, `concurrency.Session` in the clients is closed.
But some clientv3 functions in the clients are not canceled. This behavior cause hangs on the clients. 

So we need to check the closing of `concurrency.Session`. And cancel contexts if the session is closed.

In CKE server, there are 2 probrems.

1. `election.Campaign` in standby CKE servers hang.
2. The leader server cannot detect the loss of his leadership.

This PR will monitor the closing of `concurrency.Session` and terminate the CKE process.
This will solve the above problems.

### DIFF
--- a/server/control.go
+++ b/server/control.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/coreos/etcd/clientv3"
 	"github.com/coreos/etcd/clientv3/concurrency"
 	"github.com/cybozu-go/cke"
 	"github.com/cybozu-go/cke/metrics"
@@ -35,6 +36,8 @@ func NewController(s *concurrency.Session, interval, gcInterval, timeout time.Du
 
 // Run execute procedures with leader elections
 func (c Controller) Run(ctx context.Context) error {
+	metrics.UpdateLeader(false)
+
 	hostname, err := os.Hostname()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


binoue (approved):
LGTM

server/control.go
```diff
@@ -97,6 +111,32 @@ func (c Controller) runLoop(ctx context.Context, leaderKey string) error {
...
+				if resp.Err() != nil {
+					return resp.Err()
+				}
+				for _, ev := range resp.Events {
+					if ev.Type == clientv3.EventTypeDelete {
+						return errors.New("leader key is deleted")
+					}
+				}
+			}
+		}
```
ymmt2005, line 137: This function cannot return `nil`.  In other words, this function must always fail.
Shouldn't we return `nil` in case of `ctx.Done()`? @masa213f

masa213f: @ymmt2005 
`ctx.Err()` seems to return a non-nil error, when `ctx.Done()` is closed.
So this function will not return `nil`.

https://golang.org/pkg/context/

>    // If Done is not yet closed, Err returns nil.
    // If Done is closed, Err returns a non-nil error explaining why:
    // Canceled if the context was canceled
    // or DeadlineExceeded if the context's deadline passed.
    // After Err returns a non-nil error, successive calls to Err return the same error.
    Err() error

PR #454, opened 2021-04-27
Fix release cke-tools workflow

ysksuzuki:
Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>


ysksuzuki:
@binoue 
[.dockerignore](https://github.com/cybozu-go/cke/blob/main/tools/.dockerignore) excludes unnecessary files. So we don't need to anything in particular, since `bin`, `plugins/bin` and `plugins/LICENSE` will be copied to a container image.

ysksuzuki:
Confirmed it works.
https://github.com/cybozu-go/cke/runs/2444827674?check_suite_focus=true

binoue (approved):
LGTM

PR #455, opened 2021-04-27
Update for Kubernetes 1.20

ysksuzuki:
This PR includes

- Update dependencies for Kubernetes 1.20
  - Add  `--service-account-issuer` to kube-apiserver options for [Service Account Token Volume Projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection). kube-apiserver raises the error `[service-account-issuer is a required flag, --service-account-signing-key-file and --service-account-issuer are required flags]` without it. 
- Disable PodSecurityPolicy
- Drop Docker support
- ComponentConfig for kube-proxy

Sonobuoy passed [here](https://github.com/cybozu-go/cke/actions/runs/824921488)


cluster.go
```diff
@@ -161,34 +162,67 @@ func (p SchedulerParams) MergeConfig(base *schedulerv1beta1.KubeSchedulerConfigu
...
+}
+
+// MergeConfig merges the input struct with `base`.
+func (p ProxyParams) MergeConfig(base *proxyv1alpha1.KubeProxyConfiguration) (*proxyv1alpha1.KubeProxyConfiguration, error) {
+	// FOR IMPLEMENTORS.
+	// DO NOT SUPPORT MORE THAN ONE ComponentConfig VERSIONS.
+	// When we need to upgrade the component config version, users will
+	// stop CKE, update cluster.yml in etcd, then start the new CKE.
+	// So, CKE should only support the latest config version.
+	cfg := *base
```
morimoto-cybozu, line 184: Use `DeepCopy()` not to destroy `base`.
Though the current calls of this method seem to be unaffected whether the method is destructive or not, it would be safer to keep methods non-destructive.

morimoto-cybozu (changes requested):

cluster.go
morimoto-cybozu: P.S.
`json.Unmarshal()` is the destroyer.
https://play.golang.org/p/Fo0FxXAltjC

zoetrope (dismissed):
LGTM

morimoto-cybozu (approved):
LGTM

PR #465, opened 2021-05-21
Update for etcd 3.4.16

yokaze:
Make CKE compatible for etcd 3.4.
`enable-v2: false` is omitted because it is the default for etcd 3.4.


ymmt2005 (approved):
LGTM

PR #474, opened 2021-09-02
Implement ckecli rq cancel-all

yokaze:
This PR adds `ckecli reboot-queue cancel-all` command.

### DIFF
--- a/pkg/ckecli/cmd/reboot_queue_cancel_all.go
+++ b/pkg/ckecli/cmd/reboot_queue_cancel_all.go
@@ -1,0 +1,43 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/cybozu-go/cke"
+	"github.com/cybozu-go/well"
+	"github.com/spf13/cobra"
+)
+
+var rebootQueueCancelAllCmd = &cobra.Command{
+	Use:   "cancel-all",
+	Short: "cancel all the reboot queue entries",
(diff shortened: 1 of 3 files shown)


morimoto-cybozu:
@yokaze 
Is this PR ready for review?

yokaze:
@morimoto-cybozu
Yes. I turned it to a draft for a while to check if we need a null check at `for _, entry := range entries`.

umezawatakeshi (approved):

pkg/ckecli/cmd/reboot_queue_cancel_all.go
```diff
@@ -0,0 +1,43 @@
...
+			}
+
+			for _, entry := range entries {
+				if entry.Status == cke.RebootStatusCancelled {
+					continue
+				}
+
+				entry.Status = cke.RebootStatusCancelled
+				err := storage.UpdateRebootsEntry(ctx, entry)
+				if err != nil {
```
morimoto-cybozu, line 30: `ErrNotFound` would be returned if the queue entry has just finished.  Though we can retry `ckecli rq cancel-all`, it would be good to ignore `ErrNotFound` here.

morimoto-cybozu (changes requested):

morimoto-cybozu (approved):
LGTM

PR #482, opened 2021-09-29
Delete non-protected pod when Eviction API returns any kind of errors

masa213f:
When there is misconfiguration of PodDisruptionBudgets, the Eviction API returns a 500 Internal Server Error response.
Even in this case, delete non-protected pods.

### DIFF
--- a/op/reboot.go
+++ b/op/reboot.go
@@ -293,7 +293,12 @@ func evictOrDeleteNodePod(ctx context.Context, cs *kubernetes.Clientset, n *cke.
 		})
 		switch {
 		case apierrors.IsNotFound(err):
-		case apierrors.IsTooManyRequests(err) && !protected[pod.Namespace]:
+		case err != nil && !protected[pod.Namespace]:
+			log.Warn("failed to evict non-protected pod", map[string]interface{}{
+				"namespace": pod.Namespace,
+				"name":      pod.Name,
+				log.FnError: err,
+			})
 			err := cs.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{})
 			if err != nil {
 				return nil, err
--- a/op/reboot.go
+++ b/op/reboot.go
@@ -3,7 +3,6 @@ package op
 import (
 	"bytes"
 	"context"
-	"errors"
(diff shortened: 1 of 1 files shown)


op/reboot.go
```diff
@@ -293,7 +291,12 @@ func evictOrDeleteNodePod(ctx context.Context, cs *kubernetes.Clientset, n *cke.
 		})
 		switch {
 		case apierrors.IsNotFound(err):
-		case apierrors.IsTooManyRequests(err) && !protected[pod.Namespace]:
+		case err != nil && !protected[pod.Namespace]:
+			log.Warn("failed to evict non-protected pod", map[string]interface{}{
+				"namespace": pod.Namespace,
+				"name":      pod.Name,
+				log.FnError: err,
+			})
```
zoetrope, line 299: The same kind of log appears to be output twice 🤔

zoetrope (approved):
LGTM

op/reboot.go
```diff
@@ -250,14 +249,13 @@ func listProtectedNamespaces(ctx context.Context, cs *kubernetes.Clientset, ls *
 	return nss, nil
 }
 
-func existJobManagedPod(ctx context.Context, cs *kubernetes.Clientset, n *cke.Node) (bool, error) {
+func verifyJobManagedPod(ctx context.Context, cs *kubernetes.Clientset, n *cke.Node) error {
```
yokaze, line 252:
```suggestion
func checkJobPodNotExist(ctx context.Context, cs *kubernetes.Clientset, n *cke.Node) error {
```

op/reboot.go
```diff
@@ -356,11 +359,9 @@ func (c drainCommand) Run(ctx context.Context, inf cke.Infrastructure, _ string)
 	}
 
 	for _, n := range c.nodes {
-		exists, err := existJobManagedPod(ctx, cs, n)
+		err := verifyJobManagedPod(ctx, cs, n)
```
yokaze, line 362:
```suggestion
		err := checkJobPodNotExist(ctx, cs, n)
```

yokaze (changes requested):

yokaze (approved):
LGTM

PR #496, opened 2021-10-11
Use policy/v1 PodDisruptionBudget instead of policy/v1beta1

morimoto-cybozu:

### DIFF
--- a/mtest/reboot-deployment.yaml
+++ b/mtest/reboot-deployment.yaml
@@ -17,7 +17,7 @@ spec:
         - name: ubuntu
           image: quay.io/cybozu/testhttpd:0
 ---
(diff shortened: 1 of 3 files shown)


umezawatakeshi:
Do we have to update PDBs at https://github.com/cybozu-go/cke/blob/bdf7f08e18bb14fec0f4f6c20c74ff6cd1131cb5/example/kube-flannel.yaml#L2 and https://github.com/cybozu-go/cke/blob/bdf7f08e18bb14fec0f4f6c20c74ff6cd1131cb5/op/reboot.go#L16 ?

umezawatakeshi:
I see. :+1:

Issue #484, opened 2021-09-30
Switch from no-longer-maintained gopass


morimoto-cybozu:
## What

CKE uses [gopass](https://github.com/howeyc/gopass), which is no longer maintained.
Find and use an alternative.

## How

[terminal](https://pkg.go.dev/golang.org/x/crypto/ssh/terminal) is suggested. -> terminal is deprecated and moved to [term](https://pkg.go.dev/golang.org/x/term).

## Checklist

- [ ] Finish implentation of the issue
- [ ] Test all functions
- [ ] Have enough logs to trace activities
- [ ] Notify developers of necessary actions

PR #511, opened 2021-11-09
use term package instead of gopass

umezawatakeshi:
part of https://github.com/cybozu-go/cke/issues/484

### DIFF
--- a/pkg/ckecli/cmd/vault_init.go
+++ b/pkg/ckecli/cmd/vault_init.go
@@ -10,8 +10,8 @@ import (
 	"github.com/cybozu-go/cke"
 	"github.com/cybozu-go/well"
 	vault "github.com/hashicorp/vault/api"
-	"github.com/howeyc/gopass"
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/ckecli/cmd/vault_init.go
```diff
@@ -59,6 +60,30 @@ path "cke/*"
...
+	screen := struct {
+		io.Reader
+		io.Writer
+	}{os.Stdin, os.Stdout}
+
+	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
+	if err != nil {
+		return "", err
+	}
+	defer term.Restore(0, oldState)
```
morimoto-cybozu, line 77: The godoc recommends not to use 0.
https://pkg.go.dev/golang.org/x/term#pkg-overview

```suggestion
	defer term.Restore(int(os.Stdin.Fd()), oldState)
```

pkg/ckecli/cmd/vault_init.go
```diff
@@ -59,6 +60,30 @@ path "cke/*"
...
+
+	t := term.NewTerminal(screen, "")
+	p, err := t.ReadPassword(prompt)
```
morimoto-cybozu, line 80: A new terminal does not seem necessary.
I'm not sure whether `MakeRaw()` is necessary or not.

```suggestion
	fmt.Print(prompt)
	p, err := term.ReadPassword(int(os.Stdin.Fd()))
```

morimoto-cybozu (changes requested):

pkg/ckecli/cmd/vault_init.go
umezawatakeshi: according to my local poc just now, we have to print '\n' after ReadPassword. i.e. the following is enough:
```
	fmt.Print(prompt)
	p, err := term.ReadPassword(int(os.Stdin.Fd()))
	fmt.Println()
```

`MakeRaw` is required if we use `t.ReadPassword` but it does not seems required if we use `term.ReadPassword`.

morimoto-cybozu (approved):
LGTM

PR #512, opened 2021-11-12
Update Endpoints/EndpointSlices when rebooting CP Node

morimoto-cybozu:
This PR improves the CKE strategy to update `kubernetes`/`cke-etcd` Endpoints/EndpointSlices before/after rebooting a CP Node.
This PR also adds tests for the reconciliation of EndpointSlices.

### DIFF
--- a/server/strategy.go
+++ b/server/strategy.go
@@ -69,7 +69,7 @@ func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, constraints *cke.Constrain
 	}
 
 	// 7. Maintain k8s resources.
-	if ops := k8sMaintOps(c, cs, resources, nf); len(ops) > 0 {
+	if ops := k8sMaintOps(c, cs, resources, reboot, nf); len(ops) > 0 {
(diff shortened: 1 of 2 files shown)


yokaze (changes requested):
I suspect it would be better to have a new test case, that ensures to perform `update-endpointslice` instead of `reboot` in case of rebooting control-plane nodes.

morimoto-cybozu:
@yokaze 
Sure, I'll add it.

In my intention the test case you suggested was expressed in `(Master|Etcd)(Endpoints|EndpointSlice)Update*` and `EndpointsWithRebootEntry`.
The case `*Update*` shows that CKE executes `update-endpoints` or `update-endpointslice` if the observed status of Endpoints or EndpointSlice differs from the expected one.
And the case `EndpointsWithRebootEntry` shows the expected status for the cluster with a reboot-queue entry, which differs from the normal status.

I have overlooked that these test cases do not guarantee that `update-*` runs *before* processing the reboot queue.

yokaze (approved):
LGTM

PR #514, opened 2021-11-23
Use the right method to validate annotations

ymmt2005:
Kubernetes 1.21 provides a validation function for annotations
in k8s.io/apimachinery/pkg/api/validation package.

https://pkg.go.dev/k8s.io/apimachinery/pkg/api/validation#ValidateAnnotations


zoetrope (approved):
LGTM

PR #533, opened 2022-04-19
Allows specifying the time to reboot.

kmdkuk:
Signed-off-by: kouki <kouworld0123@gmail.com>


kmdkuk:
Better to run `ckecli rq disable/enable` periodically.

PR #540, opened 2022-06-20
parallel reboot feature

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/2010


docs/cluster.md
```diff
@@ -65,21 +65,21 @@ Taint
...
+| `boot_check_command`       | true     | array                            | A command to check nodes booted.  List of strings.                      |
+| `eviction_timeout_seconds` | false    | *int                             | Deadline for eviction. Must be positive. Default: 600 (10 minutes).     |
+| `command_timeout_seconds`  | false    | *int                             | Deadline for rebooting. Zero means infinity. Default: wait indefinitely |
+| `max_concurrent_reboots`   | false    | *int                             | Maximum number of nodes to be rebooted concurrently. Default: 1         |
+| `protected_namespaces`     | false    | [`LabelSelector`][LabelSelector] | A label selector to protect namespaces.                                 |
+
+`reboot_command` is the command to reboot a node. The node is passed as a command argument.
+The command should return zero if the reboot is successfully started.
+
+`boot_check_command` is the command to check a node booted. The node and the unix time when the reboot command is run is passed as a command argument.
```
zoetrope, line 80:
```suggestion
`boot_check_command` is the command to check a node booted. The node and the unix time when the reboot command is run are passed as a command argument.
```

server/watch.go
```diff
@@ -46,6 +47,7 @@ func startWatcher(ctx context.Context, etcd *clientv3.Client, ch chan<- struct{}
 			key := string(ev.Kv.Key)
 			switch {
 			case key == cke.KeyCluster || strings.HasPrefix(key, cke.KeyResourcePrefix):
+				fmt.Println("******** detected cluster update in etcd ********")
```
zoetrope, line 50: print debug?

server/strategy.go
```diff
@@ -16,7 +16,7 @@ import (
 
 // DecideOps returns the next operations to do and the operation phase.
 // This returns nil when no operations need to be done.
-func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, constraints *cke.Constraints, resources []cke.ResourceDefinition, reboot *cke.RebootQueueEntry) ([]cke.Operator, cke.OperationPhase) {
+func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, constraints *cke.Constraints, resources []cke.ResourceDefinition, rqEntries []*cke.RebootQueueEntry, newlyDrained []*cke.RebootQueueEntry, drainCompleted []*cke.RebootQueueEntry, drainTimedout []*cke.RebootQueueEntry, rebootDequeued []*cke.RebootQueueEntry) ([]cke.Operator, cke.OperationPhase) {
```
zoetrope, line 19: I think there are too many arguments of DecideOps.

Can you put the arguments together into a structure?

op/reboot.go
```diff
@@ -1,15 +1,15 @@
 package op
```
zoetrope, line 1: I think this file is a little large.
Can you create a directory like `etcd` and split the file?

server/strategy.go
```diff
@@ -625,48 +641,63 @@ func cleanOps(c *cke.Cluster, nf *NodeFilter) (ops []cke.Operator) {
...
-	if entry == nil {
+func rebootOps(c *cke.Cluster, rqEntries []*cke.RebootQueueEntry, newlyDrained []*cke.RebootQueueEntry, drainCompleted []*cke.RebootQueueEntry, drainTimedout []*cke.RebootQueueEntry, rebootDequeued []*cke.RebootQueueEntry, nf *NodeFilter) (ops []cke.Operator) {
+	if len(rqEntries) == 0 {
 		return nil
 	}
-	if entry.Status == cke.RebootStatusCancelled {
-		return []cke.Operator{op.RebootDequeueOp(entry.Index)}
-	}
-	if len(c.Reboot.Command) == 0 {
+	if len(c.Reboot.RebootCommand) == 0 {
```
zoetrope, line 648: I suggest checking the settings with `validateReboot`.
Because we will notice the problem earlier.

https://github.com/cybozu-go/cke/blob/0101d79b51f6662d91f98159e4f161d55339f015/cluster.go#L460

op/reboot.go
```diff
@@ -249,48 +551,75 @@ func listProtectedNamespaces(ctx context.Context, cs *kubernetes.Clientset, ls *
...
+		if owner != nil {
+			switch owner.Kind {
+			case "DaemonSet":
+				continue
+			case "Job":
+				switch pod.Status.Phase {
+				case corev1.PodPending:
+				case corev1.PodSucceeded:
+				case corev1.PodFailed:
+				default:
```
zoetrope, line 579: It is not clear what `default` indicates.
`Running` and `Unknown`?

server/strategy.go
```diff
@@ -79,12 +79,12 @@ func DecideOps(c *cke.Cluster, cs *cke.ClusterStatus, constraints *cke.Constrain
...
 	// 9. Uncordon nodes if nodes are cordoned by CKE.
-	if o := rebootUncordonOp(nf); o != nil {
+	if o := rebootUncordonOp(rqEntries, nf); o != nil {
 		return []cke.Operator{o}, cke.PhaseUncordonNodes
 	}
 
 	// 10. Reboot nodes if reboot request has been arrived to the reboot queue, and the number of unreachable nodes is less than a threshold.
-	if ops := rebootOps(c, reboot, nf); len(ops) > 0 {
+	if ops := rebootOps(c, rqEntries, newlyDrained, drainCompleted, drainTimedout, rebootDequeued, nf); len(ops) > 0 {
 		if len(nf.SSHNotConnectedNodes(nf.cluster.Nodes, true, true)) > constraints.RebootMaximumUnreachable {
```
zoetrope, line 88: Is this condition correct?
Rebooting nodes cannot have ssh connections, so should they be excluded?

server/strategy.go
umezawatakeshi: - If the cluster administrator does not need rebooting support, those fields are empty.
- During CKE upgrade, those fields may be empty.

So, checking those fields in `validateReboot` may cause a problem.

server/strategy.go
umezawatakeshi: Hmm... the behavior of the point seems to need more consideration.

server/strategy.go
umezawatakeshi: Which arguments are to be put together? All reboot queue entry slices (5 arguments) ?

op/reboot.go
umezawatakeshi: Yes. there "case"s are whitelist for silent discard.

op/reboot.go
umezawatakeshi: If the problem is that this file is large, why not simply divide the file?

op/reboot.go
```diff
@@ -299,143 +628,188 @@ func evictOrDeleteNodePod(ctx context.Context, cs *kubernetes.Clientset, n *cke.
...
+			continue
 		}
-	}
 
-	var targets []*corev1.Pod
-	for _, n := range c.nodes {
-		pods, err := evictOrDeleteNodePod(ctx, cs, n, protected)
-		if err != nil {
-			return err
+		err = checkPodDeletion(ctx, cs, entry.Node, protected)
```
morimoto-cybozu, line 754: This seems somewhat dangerous because a named return value `err` is used.
This `err` may be protected by `err := ...` above, but such a protection is fragile.
I suggest not to use named return values because we return the successful results just at the end.

In general, I prefer to use `err := ` instead of `err = ` in a new block to not overwrite the outside `err`.
`err = ` expresses a will of overwriting even when the value is consumed only in the inner block.
It would be confusing.

morimoto-cybozu (changes requested):
Still reviewing...

op/reboot.go
```diff
@@ -18,96 +18,336 @@ import (
...
+			if err != nil {
+				return err
+			}
+
+			err = checkJobPodNotExist(ctx, cs, entry.Node)
+			if err != nil {
+				return err
+			}
+
+			_, err = nodesAPI.Patch(ctx, entry.Node, types.StrategicMergePatchType, []byte(`
```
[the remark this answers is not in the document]
morimoto-cybozu: If the check uses cached information, even the "cordon and check" steps can overlook a new Job Pod.
As described above this oversight would be acceptable.
It would be good to comment about this.

The comment "This should not happen" in `checkPodDeletion()` seems too strong.

mtest/operators_test.go
```diff
@@ -91,273 +91,599 @@ func nodesShouldBeSchedulable(nodes ...string) {
...
+
+		cluster.Reboot.BootCheckCommand = originalBootCheckCommand
+		_, err = ckecliClusterSet(cluster)
+		Expect(err).ShouldNot(HaveOccurred())
+		waitRebootCompletion(cluster)
+
+		By("ckecli reboot-queue disable disables reboot queue processing")
+		ckecliSafe("reboot-queue", "disable")
+		// wait for the previous reconciliation to be done
+		time.Sleep(time.Second * 3)
```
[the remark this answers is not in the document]
umezawatakeshi: I will remove the Sleep.

mtest/operators_test.go
```diff
@@ -91,273 +91,599 @@ func nodesShouldBeSchedulable(nodes ...string) {
...
+
+		By("Reboot operation will protect all pods if protected_namespaces is nil")
+		nodeName := deploymentPods.Items[0].Spec.NodeName
+		_, _, err = ckecliWithInput([]byte(nodeName), "reboot-queue", "add", "-")
+		Expect(err).ShouldNot(HaveOccurred())
+		currentWriteIndex += 1
+		rebootShouldNotProceed()
+
+		env := well.NewEnvironment(context.Background())
+		env.Go(func(ctx context.Context) error {
```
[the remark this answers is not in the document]
umezawatakeshi: Indeed... I wonder why the `ckecli()` is wrapped by goroutine too.

op/reboot.go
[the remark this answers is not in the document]
umezawatakeshi: There is a race as you pointed out, but I think it's acceptable. I will add a comment.

> The comment "This should not happen" in checkPodDeletion() seems too strong.

I agree.

op/reboot.go
```diff
@@ -299,143 +628,188 @@ func evictOrDeleteNodePod(ctx context.Context, cs *kubernetes.Clientset, n *cke.
...
+	}
+
+	cs, err := inf.K8sClient(ctx, apiserver)
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 
-	protected, err := listProtectedNamespaces(ctx, cs, c.protectedNamespaces)
+	protected, err := listProtectedNamespaces(ctx, cs, c.Reboot.ProtectedNamespaces)
```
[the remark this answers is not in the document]
umezawatakeshi: Indeed.
Drain completion check should not be affected by whether Pod NSs are protected or not. It's not necessary to pass `protected` slice to `checkPodDeletion()` nor calculate it.

op/reboot.go
umezawatakeshi: Agree. I modify it not to use named return value.

server/strategy.go
zoetrope: Arguments for reboot entry. (5 args)

op/reboot.go
zoetrope: Yes, I think so.

server/strategy.go
zoetrope: I see.

zoetrope (approved):
LGTM

server/control.go
```diff
@@ -311,23 +312,43 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
...
+		if disabled {
+			rqEntries = nil
 		}
 	}
-	ops, phase := DecideOps(cluster, status, constraints, rcs, reboot)
+
+	nf := NewNodeFilter(cluster, status)
+	apiServers := map[string]bool{}
+	for _, node := range cluster.Nodes {
+		if nf.nodeStatus(node).APIServer.Running {
```
morimoto-cybozu, line 337: I'm not sure whether we should check the running API server or the expected API server, i.e. `nf.ControlPlane()`.

op/reboot_decide.go
```diff
@@ -0,0 +1,265 @@
...
+// It calls podHandler for each Pods not owned by Job nor DaemonSet and calls jobPodHandler for each running Pods owned by a Job.
+// If those handlers returns error, this function returns the error immediately.
+func enumeratePods(ctx context.Context, cs *kubernetes.Clientset, node string,
+	podHandler func(pod *corev1.Pod) error, jobPodHandler func(pod *corev1.Pod) error) error {
+
+	podList, err := cs.CoreV1().Pods(corev1.NamespaceAll).List(ctx, metav1.ListOptions{
+		FieldSelector: fields.SelectorFromSet(fields.Set{"spec.nodeName": node}).String(),
+	})
+	if err != nil {
+		return err
```
morimoto-cybozu, line 30: The error handling of `enumeratePods()` seems somewhat dangerous because the function mixes the real errors and the statuses of unsatisfied conditions.
The callers of this function cannot distinguish the real errors from condition-unsatisfied situations.

reboot.go
```diff
@@ -1,27 +1,75 @@
...
+	}
+	return ret
+}
+
+func CountRebootQueueEntries(entries []*RebootQueueEntry) map[string]int {
+	ret := map[string]int{}
+	ret[string(RebootStatusQueued)] = 0
+	ret[string(RebootStatusDraining)] = 0
+	ret[string(RebootStatusRebooting)] = 0
+	ret[string(RebootStatusCancelled)] = 0
```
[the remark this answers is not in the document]
umezawatakeshi: It is required to output zero value metrics even if there are no entries with the corresponding status.

mtest/reboot_test.go
```diff
@@ -0,0 +1,762 @@
...
+			if deploy.Status.ReadyReplicas != 10 {
+				return fmt.Errorf("deployment is not ready")
+			}
+			return nil
+		}).Should(Succeed())
+
+		By("Starting to reboot nodes")
+		// worker nodes first, then API servers.
+		rebootQueueAdd(workerNodeSlice)
+		rebootQueueAdd(apiServerSlice)
```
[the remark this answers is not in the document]
umezawatakeshi: I will disable/enable it.

server/control.go
umezawatakeshi: In this case, "expected API servers" should be used.

`apiServers` slice is used to check whether any of API servers are rebooting. If an API server is rebooting, "running API servers" might not contain the server. It will cause miss-detection.

So, I will fix the line.

morimoto-cybozu (approved):
LGTM

PR #544, opened 2022-07-05
gitignore some new artifacts

umezawatakeshi:

### DIFF
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,3 +1,5 @@
 cke
+cke-localproxy
 ckecli
+compile_resources
 LICENSE


docker/.gitignore
```diff
@@ -1,3 +1,5 @@
 cke
+cke-localproxy
 ckecli
+compile_resources
```
morimoto-cybozu, line 4: `compile_resources` is built by `make install` indeed but not used in the form of a file.
Installing `compile_resources` as a part of `make install` seems excessive.

morimoto-cybozu (approved):
I've left a comment, but approve the change.

PR #549, opened 2022-07-19
add inter-node distribution for cluster-dns

umezawatakeshi:
part of https://github.com/cybozu-go/neco/issues/2036


masa213f (approved):
LGTM

PR #551, opened 2022-07-26
sonobuoy test: follow flatcar 3227.2.0

umezawatakeshi:
The GCE image of flatcar 3227.2.0 (the latest version when this PR is submitted) has different resolv.conf content from older versions.
CKE cannot bootstrap worker nodes because `docker pull` from public repo fails on worker nodes due to the difference.


kmdkuk (approved):
LGTM

Issue #545, opened 2022-07-07
Kubernetes 1.23


masa213f:
## What

Upgrade Kubernetes version to 1.23

## How

- Upgrade container images, including etcd
- CoreDNS, Kubernetes, cke-tools including CNI plugins, etcd, pause, unbound
- Update containerd used in mtest
- Update sonobuoy

## TODO

- The Kube-scheduler's `--port` option is deplicated. Delete it.
https://github.com/cybozu-go/cke/blob/release-1.21/op/k8s/scheduler_boot.go#L111

- The Kubelet's `--register-with-taints` option  is deplicated. Update the following document.
https://github.com/cybozu-go/cke/blob/release-1.21/docs/cluster.md#boot-taints

- `kubescheduler.config.k8s.io/v1beta1` will be deleted. Use `v1beta3`.
https://github.com/cybozu-go/cke/blob/release-1.21/mtest/cke-cluster.yml#L27
https://github.com/cybozu-go/cke/blob/release-1.21/testdata/cluster.yaml#L43

## Checklist

- [x] Finish implentation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #557, opened 2022-08-04
Stop using deprecated --register-with-taints option

masa213f:
part of https://github.com/cybozu-go/cke/issues/545

`--register-with-taints` option is deprecated, so use `RegisterWithTaints` of KubeletConfiguration instead.

kubelet uses the `RegisterWithTaints` field only when initializing Node resources.
After registration, this field does not affect Node resources.
So I fixed the kubelet_boot op to render the RegisterWithTaints at all times if `boot_taints` exists.
https://github.com/kubernetes/kubernetes/blob/v1.23.9/pkg/kubelet/kubelet_node_status.go#L311

### DIFF
--- a/docs/cluster.md
+++ b/docs/cluster.md
@@ -211,7 +211,7 @@ Fields in the below table have default values:
 
 `TLSCertFile`, `TLSPrivateKeyFile`, `Authentication`, `Authorization`, and `ClusterDNS` are managed by CKE and are not configurable.
 `RegisterWithTaints` is managed by CKE when `boot_taints` exists in KubeletParams.
-When same taints are specified in both `boot_taints` (KubeletParams) and `RegisterWithTaints` (KubeletConfiguration), CKE respects `boot_taints`.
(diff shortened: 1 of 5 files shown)


docs/cluster.md
```diff
@@ -210,6 +210,8 @@ Fields in the below table have default values:
 | `VolumePluginDir`       | `/opt/volume/bin` |
 
 `TLSCertFile`, `TLSPrivateKeyFile`, `Authentication`, `Authorization`, and `ClusterDNS` are managed by CKE and are not configurable.
+`RegisterWithTaints` is managed by CKE when `boot_taints` exists in KubeletParams.
+When same taints are specified in both `boot_taints` (KubeletParams) and `RegisterWithTaints` (KubeletConfiguration), CKE respects `boot_taints`.
```
morimoto-cybozu, line 214:
```suggestion
When taints with the same key are specified in both `boot_taints` (KubeletParams) and `RegisterWithTaints` (KubeletConfiguration), CKE respects `boot_taints`.
```

morimoto-cybozu (changes requested):

morimoto-cybozu (approved):
LGTM

kmdkuk (approved):

PR #565, opened 2022-08-29
Improve agent connection handling during rebooting

umezawatakeshi:
- release agent connection immediately after the node's state transition to `rebooting`
- apply dial timeout to ssh session creation
- shorten dial timeout


masa213f (approved):
LGTM

PR #568, opened 2022-09-16
Add serials for VMs

ysksuzuki:
This PR fixes an issue that pmctl2 doesn't connect each VM properly. Placemat manages VM information with a hash map where the key is serial. But the node resources defined in cluster.yaml for mtest don't contain serials.
https://github.com/cybozu-go/placemat/blob/27b37fc6a1d26ecc9ff3b03ad237701e8417eb64/v2/pkg/placemat/cluster.go#L117

The serials are calculated with sha1sum as placemat-menu does.
https://github.com/cybozu-go/neco/blob/b394a4de83c0af1d3662649ba6bb4df0348c36db/menu/cluster.go#L263

### DIFF
--- a/mtest/cluster.yml
+++ b/mtest/cluster.yml
@@ -16,6 +16,8 @@ interfaces:
 cpu: 1
 memory: 3G
 ignition: host1.ign
+smbios:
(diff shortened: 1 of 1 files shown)


yokaze (approved):
LGTM

PR #572, opened 2022-10-07
Don't use tainted node as control plane node

morimoto-cybozu:

### DIFF
--- a/sabakan/generator.go
+++ b/sabakan/generator.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/cybozu-go/cke"
+	"github.com/cybozu-go/cke/op"
 	"github.com/cybozu-go/log"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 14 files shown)


sabakan/generator.go
```diff
@@ -633,6 +653,40 @@ func (g *Generator) getUnusedMachines(usedNodes []*cke.Node) []*Machine {
...
+	n, ok := g.k8sNodeMap[m.Spec.IPv4[0]]
+	if !ok {
+		// We can return false even if the Kubernetes Node's taints are unknown.
+		// This is because the conditions on the taints are not mandatory.
+		return false
+	}
+
+OUTER:
+	for _, t := range n.Spec.Taints {
+		if strings.HasPrefix(t.Key, "node.kubernetes.io/") || t.Key == "cke.cybozu.com/state" || t.Key == op.CKETaintMaster {
```
zoetrope, line 666: I prefer to define `cke.cybozu.com/state` as const.

zoetrope (approved):

sabakan/generator.go
morimoto-cybozu: I also prefer to use constants:-)
`sabakan/generator.go` contains so many string literals that I've given up defining them as constants in this PR.

PR #584, opened 2022-11-28
Support Kubernetes v1.24

kmdkuk:
part of cybozu-go/neco#2114

- Support Kubernetes v1.24
- Update some depencencies
- Remove part of kubelet bootstrap settings related to dockershim removal
  - ref: https://github.com/kubernetes/kubernetes/pull/106907

Signed-off-by: kouki <kouworld0123@gmail.com>

PR #592, opened 2022-12-13
Mount directories related to CNI on kubelet

kmdkuk:
In #584 I removed the mount of directories related to CNI.
However, during validation with neco, I found that it was necessary to mount it on the pod with HostPath, so I'm reverting it back.

- coild: https://github.com/cybozu-go/neco/blob/ece450914fa84e9bb62203ff743815a18044203b/etc/coil.yaml#L8491-L8496
- cilium: https://github.com/cybozu-go/neco/blob/ece450914fa84e9bb62203ff743815a18044203b/etc/cilium.yaml#L1245-L1252

Signed-off-by: kouki <kouworld0123@gmail.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ This project employs a versioning scheme described in [RELEASE.md](RELEASE.md#ve
 ## [Unreleased]
 
 ### Changed
+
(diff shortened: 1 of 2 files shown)


zoetrope (approved):
LGTM

PR #593, opened 2022-12-14
Remove memory limits from cluster-dns manifests

ysksuzuki:
The spike in DNS queries causes a spike in memory usage of CoreDNS pods, which let to the pods running out of memory and getting OOM Killed. This PR removes memory limits to prevent this issue that occurring in a large cluster.

Signed-off-by: Yusuke Suzuki <yusuke-suzuki@cybozu.co.jp>


kmdkuk (approved):
LGTM

PR #599, opened 2022-12-28
Fix node filter to check etcd in-sync status properly

morimoto-cybozu:
The map `cke.EtcdClusterStatus.InSyncMembers` is filled by `GetEtcdClusterStatus()` in `op/status.go`.
It holds not only in-sync members but also not-in-sync members and its length is always equal to the length of `cke.EtcdClusterStatus.Members`.
This PR fixes a node filter for checking whether all etcd members are in-sync or not.
This PR also fixes the test of the CKE strategy by updating input data closer to the real.


zeroalphat (approved):
LGTM

PR #600, opened 2023-01-10
 Fix added-but-not-marked etcd members

morimoto-cybozu:
This PR adds a decision step of the CKE strategy to maintain half-added etcd members.
See the issue #579 for the details of "half-added members."
This PR also changes a node filter to not try to initialize half-added etcd members.

Fixes: #579 


zeroalphat (approved):
LGTM

PR #605, opened 2023-01-26
add download-container-images tool

umezawatakeshi:

### DIFF
--- a/tools/download-container-images/main.go
+++ b/tools/download-container-images/main.go
@@ -1,0 +1,57 @@
+package main
+
+/*
+ * download-container-images
+ *
+ * Pull container images in artifacts.go and save them to tar file.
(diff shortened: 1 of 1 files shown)


masa213f (changes requested):
The `tools` directory is for `cke-tools`, which is a set of small programs CKE calls on k8s nodes.
So I don't think it is suitable to place `download-container-images` code.
https://github.com/cybozu-go/cke/blob/main/tools/README.md

Could you move or make another suitable directory?
In my opnion, this tools is mainly for Neco cluster, so it is enough to put it in cybozu-go/neco.

umezawatakeshi:
The modification in this PR is incorporated to https://github.com/cybozu-go/neco/pull/2200

PR #614, opened 2023-02-24
Add buildvcs option to fix build error

zeroalphat:
When releasing 1.25 of cke, the binary generation failed and the release could not be made.
The reason for the binary generation failure was that the setting to embed vcs status was enabled.
Since there is no need to embed debug information in the binaries, disable vcs status.
https://github.com/cybozu-go/cke/actions/runs/4260149159/jobs/7414496154
zeroalphat <taichi-takemura@cybozu.co.jp>


yokaze (commented):
Could you explain why this option is required?

zeroalphat:
Added explanation to the text of the PR.
After this PR merges, I will re-tag and release v1.25.0-rc.1 again.

yokaze (approved):
LGTM

PR #617, opened 2023-03-29
Update user-defined resource applying logic

terassyi:
This PR adds two features described below.

1. We can customize the order of applying resources.
2. When applying Deployment and Daemonset resources, cke waits to complete the creation or update.

We can specify the resource application order using the `cke.cybozu.com/rank` annotation. 
Please see [docs/user-resources.md](https://github.com/cybozu-go/cke/pull/617/files#diff-6bbb0727cc2dd5e0e6ed4fdc2ed3af852e555c27972e71d1f3ae052ccda095ed) for detail.

And When we create or update Deployment or DaemonSet resources, cke checks if the update or creation is completed.
If it is not completed, cke runs the `nopOp` that does nothing to wait.

### DIFF
--- a/op/status.go
+++ b/op/status.go
@@ -441,13 +441,21 @@ func GetKubernetesClusterStatus(ctx context.Context, inf cke.Infrastructure, n *
 			if err != nil {
 				return cke.KubernetesClusterStatus{}, err
 			}
+			updated, _, err := unstructured.NestedInt64(obj.UnstructuredContent(), "status", "updatedNumberScheduled")
+			if err != nil {
+				return cke.KubernetesClusterStatus{}, err
+			}
 			available, _, err := unstructured.NestedInt64(obj.UnstructuredContent(), "status", "numberAvailable")
 			if err != nil {
 				return cke.KubernetesClusterStatus{}, err
 			}
-			s.SetResourceStatus(res.Key, obj.GetAnnotations(), len(obj.GetManagedFields()) != 0, int(desired), int(available))
+			s.SetResourceStatus(res.Key, obj.GetAnnotations(), len(obj.GetManagedFields()) != 0, objStatus(desired, updated, available))
 		} else if obj.GroupVersionKind().Kind == "Deployment" {
 			desired, _, err := unstructured.NestedInt64(obj.UnstructuredContent(), "status", "readyReplicas")
+			if err != nil {
+				return cke.KubernetesClusterStatus{}, err
+			}
+			updated, _, err := unstructured.NestedInt64(obj.UnstructuredContent(), "status", "updatedReplicas")
 			if err != nil {
 				return cke.KubernetesClusterStatus{}, err
 			}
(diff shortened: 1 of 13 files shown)


storage.go
```diff
@@ -562,13 +563,31 @@ func (s Storage) GetAllResources(ctx context.Context) ([]ResourceDefinition, err
 		default:
 			return nil, errors.New("invalid resource key: " + key)
 		}
+		obj := struct {
+			Metadata struct {
+				Annotations struct {
+					Revision int64  `json:"cke.cybozu.com/revision,string"`
+					Image    string `json:"cke.cybozu.com/image"`
+					Rank     uint32 `json:"cke.cybozu.com/rank"`
```
terassyi, line 571: annotation の値は文字列で入れないといけない

resource.go
```diff
@@ -266,45 +326,13 @@ func (d ResourceDefinition) NeedUpdate(rs *ResourceStatus) bool {
...
-
 // SortResources sort resources as defined order of creation.
 func SortResources(res []ResourceDefinition) {
 	less := func(i, j int) bool {
 		a := res[i]
 		b := res[j]
-		aRank := a.rank()
-		bRank := b.rank()
+		aRank := a.Rank
+		bRank := b.Rank
```
yokaze, line 336: This change makes `Rank` a mandatory field. I suspect we need to update `/static/resources.go` and other stuffs?
https://github.com/cybozu-go/cke/blob/main/static/resources.go

Or we can use zero as a default value and resolve it here dynamically.
```go
aRank := a.Rank
if aRank == 0 {
  aRank = DecideRank(...)
}
```

server/strategy.go
```diff
@@ -582,23 +582,46 @@ func decideEtcdServiceOps(apiServer *cke.Node, svc *corev1.Service) cke.Operator
...
 	}
 	for _, res := range resources {
 		if res.Kind == cke.KindDeployment && !isReady {
 			continue
 		}
 		status, ok := ks.ResourceStatuses[res.Key]
-		if !ok || res.NeedUpdate(&status) {
+		if !ok {
 			ops = append(ops, op.ResourceApplyOp(apiServer, res, !status.HasBeenSSA))
+			// To avoid applying subsequent resources not to wait the completion the update or creation
```
yokaze, line 605: Suggestions for expression:

- This sentence is hard to read, please refrain from using "not" as possible and use words itself describing negative meaning (e.g. avoid, prevent, skip).
- It would be better to mention create and update in a chronological order.
- The word _update_ here looks like a verb. It looks difficult to use the noun form, I would like to suggest using the verb form.

```suggestion
			// Skip waiting the completion to create or update to avoid applying subsequent resources
```

yokaze (changes requested):

resource.go
```diff
@@ -38,6 +39,22 @@ const (
 	KindValidatingWebhookConfiguration = "ValidatingWebhookConfiguration"
 )
 
+// rank
+const (
+	RankNamespace                      = 10
+	RankServiceAccount                 = 20
```
zoetrope, line 45: ServiceAccount is a namespaced resource.
Is this correct?

resource.go
terassyi: About static resources, we generate them by `compile_resources` in advance of starting cke.
And its output is sorted.
So we don't have to resolve it dynamically.
But I think that adding the rank field to the output of `compile_resources` is good.

terassyi:
@yokaze 
Thank you for your review!
I fixed some sentences based on your advice.
And I added some comments.
Please review again.

resource.go
terassyi: Yes, I understand this.
I don't know why the rank of ServiceAccount is set to this value.
But, I keep the existing implementation.

yokaze: @terassyi Looks good if there were a comment?

terassyi: I added a comment.

op/status.go
```diff
@@ -472,9 +489,12 @@ func GetKubernetesClusterStatus(ctx context.Context, inf cke.Infrastructure, n *
 	return s, nil
 }
 
-func objStatus(desired, updated, available int64) bool {
-	// If we takes the status immediately after applying the resource, the value of desired may be 0.
-	// In this case, we need return false.
+func objStatus(generation, observedGeneration, desired, updated, available int64) bool {
+	if generation < observedGeneration {
```
yokaze, line 493: It should be a problem if `generation` precedes `observedGeneration`?

```suggestion
	if generation > observedGeneration {
```

terassyi: Thank you.
I fixed this.

yokaze (approved):
LGTM!

zoetrope (approved):
LGTM

ysksuzuki (approved):

yokaze (approved):

ysksuzuki (approved):

zoetrope (approved):

PR #624, opened 2023-05-02
Update for Kubernetes 1.25.9

zeroalphat:
- Update for Kubernetes 1.25.9
  - Update Kubernetes to v1.25.9
  - Update some dependencies


static/resources.go
```diff
@@ -60,17 +60,17 @@ var Resources = []cke.ResourceDefinition{
 		Namespace:  "kube-system",
 		Name:       "node-dns",
 		Revision:   4,
```
zoetrope, line 62: Don't you need to update the revision?

zoetrope: If you don't update the revision, pod may not be updated.

zeroalphat: It seemed that if the container image was being updated, the Pod would be updated.
Since the unbound and unbound_exporter images are being updated, I would expect the Pod updates to be performed.
https://github.com/cybozu-go/cke/blob/1593fe655ddd0282bfc11b8b2b2119cbabedb5fd/resource.go#L266

zoetrope (approved):

PR #627, opened 2023-05-11
Fix Calico manifest to latest version

zeroalphat:
The Calico manifest currently used in CKE testing is from the master branch and contains pre-release changes.
Since CKE does not need to use the pre-relase manifest of Calico, I have changed the file to use the manifest of the latest version of Calico.


morimoto-cybozu (approved):
LGTM

PR #630, opened 2023-05-29
Limit the number of components (kubelet, kube-proxy and rivers) that can be updated simultaneously

zoetrope:
Add a new cli option `max-concurrent-updates`.
It can limit the number of components (kubelet, kube-proxy and rivers) that can be updated simultaneously.

This change will solve the following problems:

- Prevent large amounts access to the Internet
- Prevent large amounts access to APIServer
- Prevent all kubelets from going down

Signed-off-by: zoetrope <a.ikezoe@gmail.com>

### DIFF
--- a/pkg/cke/main.go
+++ b/pkg/cke/main.go
@@ -19,12 +19,13 @@ import (
 )
 
 var (
-	flgHTTP            = pflag.String("http", "0.0.0.0:10180", "<Listen IP>:<Port number>")
-	flgConfigPath      = pflag.String("config", "/etc/cke/config.yml", "configuration file path")
-	flgInterval        = pflag.String("interval", "1m", "check interval")
-	flgCertsGCInterval = pflag.String("certs-gc-interval", "1h", "tidy interval for expired certificates")
-	flgSessionTTL      = pflag.String("session-ttl", "60s", "leader session's TTL")
-	flgDebugSabakan    = pflag.Bool("debug-sabakan", false, "debug sabakan integration")
+	flgHTTP                         = pflag.String("http", "0.0.0.0:10180", "<Listen IP>:<Port number>")
+	flgConfigPath                   = pflag.String("config", "/etc/cke/config.yml", "configuration file path")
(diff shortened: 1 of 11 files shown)


pkg/cke/main.go
```diff
@@ -19,12 +20,13 @@ import (
...
-	flgCertsGCInterval = pflag.String("certs-gc-interval", "1h", "tidy interval for expired certificates")
-	flgSessionTTL      = pflag.String("session-ttl", "60s", "leader session's TTL")
-	flgDebugSabakan    = pflag.Bool("debug-sabakan", false, "debug sabakan integration")
+	flgHTTP                 = pflag.String("http", "0.0.0.0:10180", "<Listen IP>:<Port number>")
+	flgConfigPath           = pflag.String("config", "/etc/cke/config.yml", "configuration file path")
+	flgInterval             = pflag.String("interval", "1m", "check interval")
+	flgCertsGCInterval      = pflag.String("certs-gc-interval", "1h", "tidy interval for expired certificates")
+	flgSessionTTL           = pflag.String("session-ttl", "60s", "leader session's TTL")
+	flgDebugSabakan         = pflag.Bool("debug-sabakan", false, "debug sabakan integration")
+	flgMaxConcurrentUpdates = pflag.Int("max-concurrent-updates", 10, "the maximum number of components that can be updated simultaneously")
```
morimoto-cybozu, line 29:
```suggestion
	flgMaxConcurrentUpdates = pflag.Int("max-concurrent-updates", 10, "the maximum number of worker-node components that can be updated simultaneously")
```

The description above seems too simple.
It would be nice to add an explanation for how CKE uses this parameter.
How about updating the following section?

https://github.com/cybozu-go/cke/blob/v1.25.3/docs/cluster_overview.md#worker-nodes

CKE deploys each worker-node component concurrently on multiple nodes.
You can control the deployment concurrency by specifying the `max-concurrent-updates` command-line option.

morimoto-cybozu (changes requested):
The code itself seems good to me.

morimoto-cybozu (approved):
LGTM

PR #634, opened 2023-06-12
fix to apply resources that has same rank at the same time

terassyi:
I fixed the resource applying logic.
If resources have the same rank, cke applies these resources simultaneously. 

Signed-off-by: terasihma <tomoya-terashima@cybozu.co.jp>

### DIFF
--- a/server/strategy.go
+++ b/server/strategy.go
@@ -632,8 +632,9 @@ func decideResourceOps(apiServer *cke.Node, ks cke.KubernetesClusterStatus, reso
 	}
 	rank := resources[0].Rank
 
+	// resources are sorted by rank in advance
 	for _, res := range resources {
-		if res.Rank != uint32(rank) && len(ops) > 0 {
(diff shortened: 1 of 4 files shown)


server/strategy.go
```diff
@@ -625,20 +625,38 @@ func decideResourceOps(apiServer *cke.Node, ks cke.KubernetesClusterStatus, reso
...
 	}
+
+	// decide to apply user defined resources
+	if len(resources) == 0 {
+		return ops
+	}
+	rank := resources[0].Rank
+
 	for _, res := range resources {
+		if res.Rank != uint32(rank) && len(ops) > 0 {
```
ysksuzuki, line 636:
```suggestion
		if res.Rank != rank && len(ops) > 0 {
```

ysksuzuki (changes requested):
Should we comment somewhere that `decideResourceOps` assumes that `resources` are sorted by rank?

ysksuzuki (approved):

PR #641, opened 2023-06-29
Fix retry of reboot to set timeout in each retry

morimoto-cybozu:


zeroalphat (approved):
LGTM

PR #660, opened 2023-09-07
Fix `cke_node_reboot_status` metrics

zoetrope:
Currently, there is an issue with `cke_node_reboot_status` metrics not being reflected correctly in the following cases:

- When drain/uncordon is repeated
- When node additions and deletions occur with sabakan integration

This PR calculates the value of metrics each time they are collected.
As a result we will be able to get correct metrics.

### DIFF
--- a/metrics/collector.go
+++ b/metrics/collector.go
@@ -123,6 +123,7 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
 	wg.Wait()
 }
 
+// nodeMetricsCollector implements prometheus.Collector interface.
 type nodeMetricsCollector struct {
 	storage storage
 }
@@ -141,11 +142,17 @@ func (c nodeMetricsCollector) Collect(ch chan<- prometheus.Metric) {
 
 	rqEntries, err := c.storage.GetRebootsEntries(ctx)
 	if err != nil {
+		log.Error("failed to get reboots entries", map[string]interface{}{
+			log.FnError: err,
+		})
 		return
 	}
 
 	cluster, err := c.storage.GetCluster(ctx)
 	if err != nil {
+		log.Error("failed to get reboots entries", map[string]interface{}{
+			log.FnError: err,
+		})
 		return
 	}
 	itemCounts := cke.CountRebootQueueEntries(rqEntries)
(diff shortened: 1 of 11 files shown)


sabakan/generator.go
```diff
@@ -52,6 +52,8 @@ func MachineToNode(m *Machine, tmpl *cke.Node) *cke.Node {
 	n.Labels["cke.cybozu.com/rack"] = strconv.Itoa(m.Spec.Rack)
 	n.Labels["cke.cybozu.com/index-in-rack"] = strconv.Itoa(m.Spec.IndexInRack)
 	n.Labels["cke.cybozu.com/role"] = m.Spec.Role
+	n.Labels["cke.cybozu.com/retire-date"] = m.Spec.RetireDate.Format("2006-01")
+	n.Labels["cke.cybozu.com/register-date"] = m.Spec.RegisterDate.Format("2006-01")
```
morimoto-cybozu, line 56: Contaminated?
Please create another pull request with `docs/sabakan-integration.md` updated.

morimoto-cybozu (changes requested):
The new logic seems good!
I can only leave trivial comments.

sabakan/generator.go
zoetrope: It's my mistake...

metrics/collector.go
```diff
@@ -120,3 +122,68 @@ func (c collector) Collect(ch chan<- prometheus.Metric) {
...
+	if err != nil {
+		log.Error("failed to get reboots entries", map[string]interface{}{
+			log.FnError: err,
+		})
+		return
+	}
+
+	cluster, err := c.storage.GetCluster(ctx)
+	if err != nil {
+		log.Error("failed to get reboots entries", map[string]interface{}{
```
morimoto-cybozu, line 153:
```suggestion
		log.Error("failed to get cluster", map[string]interface{}{
```

morimoto-cybozu (changes requested):

morimoto-cybozu (approved):
LGTM

morimoto-cybozu (approved):

PR #661, opened 2023-09-07
Fix blocking by kubelet-restart op

zoetrope:
`kubelet-restart` op takes a long time.
If a problem occurs during `kubelet-restart` op (e.g. controller-manager crashes), the problem will not be repaired.

In this PR, after one of `kubelet-restart` op is finished, next operation is decided again.
This will be able to repair a problem in a short time.


zeroalphat (approved):
LGTM

Issue #657, opened 2023-08-15
Add ckecli rq reset-backoff command


yokaze:
## What

When reboot-queue is stuck, normally we do some manual operation and wait for the next reboot.
The waiting time can be shortened if we can reset the backoff of a reboot entry.

## How

Implement `ckecli rq reset-backoff` to reset the backoff count and time

## Checklist

- [x] Finish implentation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #667, opened 2023-10-20
add reboot queue backoff reset command

YZ775:
This PR is implementation of following issue.
- https://github.com/cybozu-go/cke/issues/657
Signed-off-by: YZ775 <yuzuki-mimura@cybozu.co.jp>

### DIFF
--- a/docs/ckecli.md
+++ b/docs/ckecli.md
@@ -47,6 +47,7 @@ $ ckecli [--config FILE] <subcommand> args...
   - [`ckecli reboot-queue list`](#ckecli-reboot-queue-list)
   - [`ckecli reboot-queue cancel INDEX`](#ckecli-reboot-queue-cancel-index)
   - [`ckecli reboot-queue cancel-all`](#ckecli-reboot-queue-cancel-all)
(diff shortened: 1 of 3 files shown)


zeroalphat (dismissed):
LGTM

docs/ckecli.md
```diff
@@ -299,6 +300,11 @@ Cancel the specified reboot queue entry.
 
 Cancel all the reboot queue entries.
 
+### `ckecli reboot-queue reset-backoff`
+
+Reset `drain_backoff_count` and `drain_backoff_expire` of the entries in reboot queue.
+Resetting these values allows CKE to try draining nodes again immediately.
```
yokaze, line 306: - It is better to use a verb that guarantees the action being actually done, instead of using _allow_.
- This command cause the retry of entire reboot sequence, not only draining.
- Try _V-ing_ means to check the result of the action. Use "try to" if it may fail.

```suggestion
Resetting these values makes CKE try to reboot nodes again immediately.
```

yokaze (changes requested):

docs/ckecli.md
YZ775: Thank you. I fixed it.

yokaze (approved):
LGTM

PR #675, opened 2023-11-24
wip

yokaze:


mtest/kubernetes_test.go
```diff
@@ -245,6 +259,12 @@ func testKubernetes() {
 			return err
 		}).Should(Succeed())
 
+		By("querying www.dnssec-failed.org using node DNS from ubuntu pod")
```
masa213f, line 262: This test is OK. 

Could you update the `By()` just before? Currently, we uses "www.cybozu.com".
`By("querying www.google.com using node DNS from ubuntu pod")`

masa213f (approved):
LGTM

PR #695, opened 2024-01-10
Adjustment network component resources

zeroalphat:
- coredns
  - revise cpu request from 100m to 50m
  - revise memory request from 70Mi to 250Mi
- unbound(node-dns)
  - set cpu request ~28m~ 50m
  - set memory request 250Mi


masa213f (approved):
LGTM

PR #699, opened 2024-02-01
fix retaint operation

YZ775:


server/strategy.go
```diff
@@ -183,7 +183,7 @@ func k8sOps(c *cke.Cluster, nf *NodeFilter, cs *cke.ClusterStatus, maxConcurrent
 		if len(nodes) < max {
 			max = len(nodes)
 		}
-		ops = append(ops, k8s.KubeletBootOp(nodes[:max], nf.KubeletStoppedRegisteredNodes(), apiServer, c.Name, c.Options.Kubelet, cs.NodeStatuses))
+		ops = append(ops, k8s.KubeletBootOp(nodes[:max], nf.SSHConnectedNodes(nf.RegisteredNodes(nodes[:max]), true, true), apiServer, c.Name, c.Options.Kubelet, cs.NodeStatuses))
```
masa213f, line 186: `nodes` is already filterd in L181. So we need not to filter again here.

```suggestion
		ops = append(ops, k8s.KubeletBootOp(nodes[:max], nf.RegisteredNodes(nodes[:max]), apiServer, c.Name, c.Options.Kubelet, cs.NodeStatuses))
```

masa213f (changes requested):

server/strategy.go
YZ775: Thank you, it's my mistake.
I fixed it.

masa213f (approved):
LGTM

PR #707, opened 2024-03-11
change CKE to proceed rebooting immediately after draining of node  is completed

YZ775:
In the current CKE, even if `max_concurrent_reboots` is larger than 1, draining is conducted by serial and rebooting is conducted after draining of all drainable node is finished.
This behavior cause long down time of first drained node.

This PR changes the CKE to move rebooting immediately after draining of each node is finished.
In addition, this PR fixes behavior of canceled reboot queue entry.

### DIFF
--- a/server/strategy.go
+++ b/server/strategy.go
@@ -885,21 +885,19 @@ func rebootOps(c *cke.Cluster, constraints *cke.Constraints, rebootArgs DecideOp
 	}
 
 	if len(rebootArgs.RebootCancelled) > 0 {
-		phaseReboot = true
 		ops = append(ops, op.RebootCancelOp(rebootArgs.RebootCancelled))
-		return ops, phaseReboot
 	}
 	if len(rebootArgs.RebootDequeued) > 0 {
-		phaseReboot = true
 		ops = append(ops, op.RebootDequeueOp(rebootArgs.RebootDequeued))
-		return ops, phaseReboot
 	}
+	if len(ops) > 0 {
+		return ops, true
+	}
+
 	if len(rebootArgs.DrainCompleted) > 0 {
-		phaseReboot = true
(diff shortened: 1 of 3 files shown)


server/strategy.go
```diff
@@ -903,18 +910,9 @@ func rebootOps(c *cke.Cluster, constraints *cke.Constraints, rebootArgs DecideOp
...
-		phaseReboot = true
 		ops = append(ops, op.RebootDrainTimeoutOp(rebootArgs.DrainTimedout))
 	}
-	if len(rebootArgs.RebootDequeued) > 0 {
-		phaseReboot = true
-		ops = append(ops, op.RebootDequeueOp(rebootArgs.RebootDequeued))
-	}
 	if len(ops) > 0 {
 		phaseReboot = true
 	}
```
masa213f, line 918: Thank you for fixing `phaseReboot`!

I checked this variable carefully, and I was convinced that this function need not return `phaseReboot`.
This function should only determine `ops` from the given information. 

The caller(`DecideOps()`) should determine the phase. And which can be determined only by the returned `ops`.

Could you change the return value of this function  and change the caller as following?
It's okay to change it in another PR.

https://github.com/cybozu-go/cke/blob/v1.28.0/server/strategy.go#L106-L113
```
	// 11. Reboot nodes if reboot request has been arrived to the reboot queue, and the number of unreachable nodes is less than a threshold.
	if ops := rebootOps(c, constraints, rebootArgs, nf); len(ops) > 0 {
		if !nf.EtcdIsGood() {
			log.Warn("cannot reboot nodes because etcd cluster is not responding and in-sync", nil)
			return nil, cke.PhaseRebootNodes
		}
		return ops, cke.PhaseRebootNodes
	}
```

masa213f (approved):
LGTM

PR #710, opened 2024-03-14
Remove Unhealthy/Unreachable taint

masa213f:
This PR removes the following taints, which are used by the Sabakan integration.

- `cke.cybozu.com/state: unhealthy` (NoSchedule)
- `cke.cybozu.com/state: unreachable` (NoSchedule)

### `cke.cybozu.com/state: unhealthy`

The taint must not be set if the machine failure is minor enough to allow the process to run.

### `cke.cybozu.com/state: unreachable`

When a node cannot communicate, the `node.kubernetes.io/unreachable` (NoExecute) taint is added.
So `cke.cybozu.com/state: unreachable` taint does not work well.


YZ775 (approved):
LGTM

PR #715, opened 2024-03-26
change priorty of reboot queue cancel

YZ775:
In the current CKE, the dequeue operation of cancelled reboot queue entry has lower priority than other operations like drain, reboot, ...etc.
This implementation causes the dequeue operation not to work correctly when we cancel an entry whose node is under operation.
This PR solves the problem by following changes.
- separating dequeue operation between cancelled entry and completed entry
- set dequeue operation of cancelled entry higher

### DIFF
--- a/op/reboot.go
+++ b/op/reboot.go
@@ -523,6 +523,25 @@ func (c rebootDequeueCommand) Command() cke.Command {
 	}
 }
 
+//
+
+type rebootCancelOp struct {
+	rebootDequeueOp
+}
+
+// RebootCancelOp returns an Operator to dequeue cancelled reboot entries.
+func RebootCancelOp(entries []*cke.RebootQueueEntry) cke.Operator {
+	return &rebootCancelOp{
+		rebootDequeueOp{
+			entries: entries,
+		},
+	}
+}
+
+func (o *rebootCancelOp) Name() string {
+	return "reboot-cancel"
+}
+
 func listProtectedNamespaces(ctx context.Context, cs *kubernetes.Clientset, ls *metav1.LabelSelector) (map[string]bool, error) {
 	selector, err := metav1.LabelSelectorAsSelector(ls)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


server/control.go
```diff
@@ -346,13 +346,15 @@ func (c Controller) runOnce(ctx context.Context, leaderKey string, tick <-chan t
 	newlyDrained := op.ChooseDrainedNodes(cluster, apiServers, rqEntries)
 	drainCompleted, drainTimedout, _ := op.CheckDrainCompletion(ctx, inf, nf.HealthyAPIServer(), cluster, rqEntries)
 	rebootDequeued := op.CheckRebootDequeue(ctx, cluster, rqEntries)
+	RebootCancelled := op.CheckRebootCancelled(ctx, cluster, rqEntries)
```
masa213f, line 349: Please use lowercase.

```suggestion
	rebootCancelled := op.CheckRebootCancelled(ctx, cluster, rqEntries)
```

server/strategy.go
```diff
@@ -883,6 +884,11 @@ func rebootOps(c *cke.Cluster, constraints *cke.Constraints, rebootArgs DecideOp
 		return nil, false
 	}
 
+	if len(rebootArgs.RebootCancelled) > 0 {
+		phaseReboot = true
+		ops = append(ops, op.RebootDequeueOp(rebootArgs.RebootCancelled))
```
masa213f, line 889: I want to change the display name of the op to `reboot-cancel.`
Is it possible?

masa213f (changes requested):

server/strategy.go
YZ775: We can change the display name by duplicating `RebootDequeueOp` and  changing `Name()` method.
The logic of Op is identical. so I created `RebootCancelOp` which has embedded `RebootDequeueOp` and override `Name()` method.

op/reboot.go
```diff
@@ -523,6 +523,25 @@ func (c rebootDequeueCommand) Command() cke.Command {
 	}
 }
 
+//
```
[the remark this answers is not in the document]
YZ775: https://github.com/cybozu-go/cke/blob/7d544a49f6821fab2911b47101c4204b427f8d20/op/reboot.go#L184
https://github.com/cybozu-go/cke/blob/7d544a49f6821fab2911b47101c4204b427f8d20/op/reboot.go#L331

There are `//` between code block of Op, so I added it same as other Op.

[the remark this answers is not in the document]
masa213f: Oh, I see.

masa213f (approved):
LGTM

PR #716, opened 2024-03-29
Embed root.hints file in unbound container

yokaze:


chez-shanpu (approved):
LGTM

PR #726, opened 2024-04-24
change backoff algorithm to exponential backoff

YZ775:


op/reboot.go
```diff
@@ -571,7 +572,12 @@ func drainBackOff(ctx context.Context, inf cke.Infrastructure, entry *cke.Reboot
 	entry.Status = cke.RebootStatusQueued
 	entry.LastTransitionTime = time.Now().Truncate(time.Second).UTC()
 	entry.DrainBackOffCount++
-	entry.DrainBackOffExpire = entry.LastTransitionTime.Add(time.Second * time.Duration(drainBackOffBaseSeconds+rand.Int63n(int64(drainBackOffBaseSeconds*entry.DrainBackOffCount))))
+	backoffDuration := time.Second * time.Duration((1<<(entry.DrainBackOffCount-1))*drainBackOffBaseSeconds+rand.Int63n(drainBackOffBaseSeconds))
+	if backoffDuration > time.Second*drainBackOffMaxSeconds {
+		backoffDuration = time.Second * time.Duration(drainBackOffMaxSeconds+rand.Int63n(drainBackOffBaseSeconds))
+	}
+	entry.DrainBackOffExpire = entry.LastTransitionTime.Add(backoffDuration)
```
masa213f, line 579: There is some waste in this calculation.

- `time.Second` is applied every time.
- jitter is calculated twice when max backoff is exceeded
   - The first calculated jitter affects nothing to the result.

How about doing the following?

```suggestion
	backoffSeconds := (1 << (entry.DrainBackOffCount - 1)) * drainBackOffBaseSeconds
	if backoffSeconds > drainBackOffMaxSeconds {
		backoffSeconds = drainBackOffMaxSeconds
	}
	entry.DrainBackOffExpire = entry.LastTransitionTime.Add(time.Second * time.Duration(backoffSeconds+rand.Int63n(drainBackOffBaseSeconds)))
```

masa213f (approved):
LGTM

PR #728, opened 2024-04-25
Fix not to change status of repair-queue's finished entry

morimoto-cybozu:
This commit is for the following situation:
* there is a finished entry in the repair queue, especially failed one
* a new entry with the same target address is queued
* the target machine becomes healthy

The results of health check are indexed by the target machine's address,
not by the entry ID.
Though health check is not executed for a finished entry, check for
another entry with the same target address can change the status of
the failed entry into "succeeded."
This commit prevents it.


zeroalphat (approved):
LGTM

PR #742, opened 2024-05-31
update etcd image to 3.5.14.1

YZ775:


umezawatakeshi (changes requested):
We should update the following too:
- go.mod and go.sum
- `ETCD_VERSION` in Makefile
- container image in github workflow config (.github/workflows/main.yaml)
- `MANAGEMENT_ETCD_VERSION` in mtest/Makefile

YZ775:
Thank you. 
I fixed that.

umezawatakeshi (approved):

PR #746, opened 2024-06-25
Delete 'OnDelete' DaemonSet pods

tkna:
This PR adds the ability to delete Daemonset pods set to `updateStrategy: OnDelete` before rebooting.


yokaze (changes requested):
How about just deleting the pods? I doubt we need to evict the DaemonSet pods.

yokaze (approved):
LGTM

PR #764, opened 2024-08-09
use docker compose v2 in appropriate way

umezawatakeshi:
GitHub Actions' ubuntu runner no longer has `docker-compose` command.
Use `docker compose` subcommand.


YZ775 (approved):
LGTM

PR #773, opened 2024-11-28
Set /sys mount of  kubelet as read-write

YZ775:
If `/sys/fs/cgroup` is set to read-only, we cannot restart kubelet.
So change mount of `/sys` to read-write.


masa213f (approved):
LGTM

PR #784, opened 2024-12-13
fix sabakan integration to handle role that is not exist in sabakan

YZ775:
Before this PR, if roles that do not exist in sabakan is written in cke-template,`chooseWorkerTmpl()` always return such role bacause count of it is always zero.
In such a situation, increseWorker does not work.
In this PR, I change `chooseWorkerTmpl()` to skip that role if it has no sabakan machines.

### DIFF
--- a/sabakan/generator.go
+++ b/sabakan/generator.go
@@ -2,6 +2,7 @@ package sabakan
 
 import (
 	"errors"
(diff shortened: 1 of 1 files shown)


masa213f (approved):
LGTM

PR #790, opened 2025-02-13
Supress dns log

chez-shanpu:
This PR supress DNS query logs for cluster-dns and node-dns.

cf.
- https://coredns.io/plugins/log/
- https://nlnetlabs.nl/documentation/unbound/unbound.conf/


yokaze (changes requested):
It looks we need to remove the `log` plugin at `op/clusterdns/clusterdns.go:20`.

yokaze (approved):
LGTM

Issue #799, opened 2025-05-30
K8s 1.32: Update for Kubernetes 1.32


masa213f:
## What

Upgrade Kubernetes version from 1.31 to 1.32.

ref https://github.com/cybozu-go/cke/issues/786

## How

- Upgrade container image versions
    - CoreDNS, Kubernetes, cke-tools including CNI plugins, etcd, pause, unbound
- Update containerd used in mtest
- Update sonobuoy
- Enable alpha Feature Gates if necessary

## Checklist

- [x] Finish implentation of the issue
- [x] Test all functions
- [x] Have enough logs to trace activities
- [x] Notify developers of necessary actions

PR #800, opened 2025-06-24
Update go modules and actions

masa213f:
part of https://github.com/cybozu-go/cke/issues/799

This PR updates Go modules and GitHub actions to prepare for supporting Kubernetes 1.32.


cluster.go
```diff
@@ -627,7 +627,7 @@ func validateOptions(opts Options) error {
 				return err
 			}
 		} else {
-			_, err = libcni.ConfFromBytes([]byte(opts.Kubelet.CNIConfFile.Content))
+			_, err = libcni.NetworkPluginConfFromBytes([]byte(opts.Kubelet.CNIConfFile.Content))
```
masa213f, line 630: `libcni.ConfFromBytes()` is deprecated.
https://github.com/containernetworking/cni/blob/v1.3.0/libcni/conf.go#L274

PR #824, opened 2025-11-12
Implement rolling update for kube-apiservers

masa213f:
The current CKE updates the kube-apiservers all at once. And this behavior causes the complete outage of the kube-apiservers for several tens of seconds when Kubernetes updates, etc.

To improve the availability of the API server, I implement rolling updates for the kube-apiservers.

### DIFF
--- a/server/strategy.go
+++ b/server/strategy.go
@@ -165,7 +165,7 @@ func apiserverOps(c *cke.Cluster, nf *NodeFilter, cs *cke.ClusterStatus) (ops []
 	//     - Updating kube-apiservers should be performed only when all kube-apiservers are healthy.
 	//     - Other k8s components should be maintained only when all kube-apiservers are *updated* and healthy, to ensure the upgrade order.
 	//       See the version skew policy: https://kubernetes.io/releases/version-skew-policy/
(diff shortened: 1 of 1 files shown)


server/strategy.go
```diff
@@ -140,16 +140,61 @@ func riversOps(c *cke.Cluster, nf *NodeFilter, maxConcurrentUpdates int) (ops []
...
+	// Waiting for all kube-apiservers to be healthy. Because...
+	// - Updating kube-apiservers should be performed only when all kube-apiservers are healthy.
+	// - Other k8s components should be maintained only when all kube-apiservers are *updated* and healthy, to ensure the upgrade order.
+	//   See the version skew policy: https://kubernetes.io/releases/version-skew-policy/
+	if len(nf.APIServerUnhealthy(nf.ControlPlaneNodes())) > 0 {
+		// Set the unhealthy kube-apiservers to NotReady.
+		if nf.HealthyAPIServer() != nil {
+			ops = append(ops, masterEndpointOps(c, cs, nf, nil)...)
+		}
+		return ops, true
```
morimoto-cybozu: This `ops` might be empty.
If this function returns `nil, true`, `k8sOps` skips subsequent decision steps and returns `nil` immediately.
This would be OK.
However, it does not seem appropriate for `DecideOps` to proceed to subsequent decision steps.

morimoto-cybozu:
Please update changelog.

PR #828, opened 2025-11-19
Refactor strategy test

masa213f:
This PR refactors `strategy_test.go`.

- Moves the initialization logic of k8s status to the appropriate location.
- Removes unused status: `ClusterStatus.Name` and `NodeStatus.Labels`.

And some minor changes.


server/strategy_test.go
```diff
@@ -135,46 +136,20 @@ func newData() testData {
...
-
 	status := &cke.ClusterStatus{
 		ConfigVersion: cke.ConfigVersion,
-		NodeStatuses:  nodeStatuses,
-		Kubernetes: cke.KubernetesClusterStatus{
-			ResourceStatuses: map[string]cke.ResourceStatus{
-				"Namespace/foo": {Annotations: map[string]string{cke.AnnotationResourceRevision: "1"}},
-			},
-			Nodes: nodeList,
-		},
```
masa213f: The Kubernetes status should not be set here.

The current status, which is created by `newData()`, means that Kubernetes nodes and user defined resources can be accessed, though Kubernetes has not been initialized.

server/strategy_test.go
```diff
@@ -135,46 +136,20 @@ func newData() testData {
...
+		// All nodes are ssh connected, but no services are running.
+		NodeStatuses: map[string]*cke.NodeStatus{
+			nodeNames[0]: {SSHConnected: true},
```
morimoto-cybozu: In real environments, other fields such as `Etcd` would also be filled in by `GetNodeStatus()`, but with the same values as the defaults.
It would be good to describe such implicit assumptions, but...
面倒ですね。

PR #831, opened 2025-11-27
Confirm EndpointSlice in mtest

masa213f:
This PR adds a test case to confirm the `kubernetes` EndpointSlice in mtest.
Additionally, adds minor improvements to the update logic for EndpointSlices.


masa213f:
@morimoto-cybozu 
Sorry, I'm withdrawing this review. I found that endpoint checking is done in other mtest. I will fix those as well.

morimoto-cybozu (approved):

PR #890, opened 2026-06-10
Check volumesInUse in CheckDrainCompletion

YZ775:


YZ775:
dctest-with-neco-feature-branch passed
https://github.com/cybozu-private/neco-apps/actions/runs/27256640202/job/80691564581

masa213f (approved):
LGTM

We may also need to check the VolumeInUse in the repair queue.
Please check this logic and update if needed.
https://github.com/cybozu-go/cke/blob/release-1.34/op/status.go#L631-L634

YZ775:
Thank you.

Checking volumesInUse in the repair queue is not necessary at this time because non-graceful node shutdown is executed just after drain in our repair setting in cke-template.

If we excecute non-graceful node shutdown, volumes on the node will be detached even if kubelet is not alive.

PR #896, opened 2026-06-16
Add feature to repair control-plane

YZ775:
In current implementation, repair of unreachable control-plane is not executed because `EtcdIsGood` returns false if one of the  etcd member lost connections.
This adds `EtcdIsGoodForRepair`, which allows repair when etcd still has quorum and out-of-sync member is on an SSH-unreachable control-plane node.

### DIFF
--- a/mtest/repair_test.go
+++ b/mtest/repair_test.go
@@ -285,4 +285,91 @@ func testRepairOperations() {
 		waitRepairSuccess()
 		nodesShouldBeSchedulable(nodeNames[1])
 	})
+
+	// A control plane failure naturally makes its etcd member out of sync.
+	// CKE must still repair the node, because the out-of-sync member is
+	// explained by the very control plane failure being repaired.
+	// This test stops a control plane node, so it must run last.
+	It("should repair a stopped control plane node whose etcd member is out of sync", func() {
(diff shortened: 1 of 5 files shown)


mtest/repair_test.go
```diff
@@ -285,4 +285,87 @@ func testRepairOperations() {
 		waitRepairSuccess()
 		nodesShouldBeSchedulable(nodeNames[1])
 	})
+
+	// The two tests above stop control plane nodes, so they must run last.
```
morimoto-cybozu:
```suggestion
	// The two tests below stop control plane nodes, so they must run last.
```

server/strategy_test.go
```diff
@@ -170,10 +170,12 @@ func newData() testData {
 		RebootQueue: cke.RebootQueueStatus{Enabled: true},
 	}
 
+	constraints := *testConstraints
```
morimoto-cybozu: The following would be a simple and straightforward approach.

```
  constraints := &cke.Constraints{
    // move definition of testConstraints here
    ...
  }
```

morimoto-cybozu (changes requested):
LGTM!
I'll just point out minor issues.

morimoto-cybozu (approved):

PR #902, opened 2026-07-07
Add audit_webhook_config to pass audit webhook config inline

zeroalphat:
This PR adds support for the audit webhook config.
CKE can set the audit webhook config and send audit logs to a remote server.

https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#webhook-backend


op/common/mkfiles.go
```diff
@@ -124,7 +139,7 @@ func (c *FilesBuilder) Run(ctx context.Context, inf cke.Infrastructure, _ string
 				data := f.dataMap[n.Address]
 				hdr := &tar.Header{
 					Name: f.name,
-					Mode: 0o644,
+					Mode: f.mode,
```
masa213f, line 142: Please verify that this mode is correctly set to the files.
At first glance, the mode of the tar header has not been reflected.
https://github.com/cybozu-go/cke/blob/81be7ae6ee3faf5caf5838a89c35e92173c2976f/tools/write_files/main.go#L62

masa213f (changes requested):

op/common/mkfiles.go
masa213f: I haven't tested this, so I apologize if my comment is incorrect.

zeroalphat: Thanks for reviewing.
You are right. I missed that `write_files` ignores the mode in the tar header, so the mode was not reflected.
I fixed it.

masa213f (approved):
LGTM

PR #903, opened 2026-07-16
Add feature to delete Job-managed Pod

YZ775:


op/reboot_decide.go
```diff
@@ -119,16 +120,24 @@ func deleteOnDeleteDaemonSetPod(ctx context.Context, cs kubernetes.Interface, no
...
 			DryRun: []string{"All"},
 		}
 	}
 
+	// LabelSelectorAsSelector(nil) returns labels.Nothing(), so Job-managed Pods are only
+	// deleted when the cluster configuration explicitly opts in with a non-nil selector.
+	deletableJobPodSelector, err := metav1.LabelSelectorAsSelector(deletableJobPodSelectorSpec)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
YZ775: `labels.Nothing()` does not match anything. So it's ok.

op/reboot_decide.go
```diff
@@ -119,16 +120,27 @@ func deleteOnDeleteDaemonSetPod(ctx context.Context, cs kubernetes.Interface, no
...
-func doEvictOrDeleteNodePod(ctx context.Context, cs kubernetes.Interface, node string, protected map[string]bool, attempts int, interval time.Duration, dry bool) error {
+func doEvictOrDeleteNodePod(ctx context.Context, cs kubernetes.Interface, node string, protected map[string]bool, deletableJobPodSelectorSpec *metav1.LabelSelector, attempts int, interval time.Duration, dry bool) error {
 	var deleteOptions *metav1.DeleteOptions
 	if dry {
 		deleteOptions = &metav1.DeleteOptions{
 			DryRun: []string{"All"},
 		}
 	}
 
+	// // LabelSelectorAsSelector(nil) returns labels.Nothing(), but set it explicitly in case that behavior changes in the future.
```
masa213f: 1. コメントに余計なスラッシュがあります。
```suggestion
	// LabelSelectorAsSelector(nil) returns labels.Nothing(), but set it explicitly in case that behavior changes in the future.
```

2. `LabelSelectorAsSelector(nil)`の動作について

LabelSelectorAsSelectorの仕様はこのコメントに書かれている通りの動作みたいですね。

https://github.com/kubernetes/apimachinery/blob/v0.36.2/pkg/apis/meta/v1/types.go#L1321-L1322
> // matchExpressions are ANDed. ***An empty label selector matches all objects. A null*
> // *label selector matches no objects.***

仕様が明記されており、その通りに実装されている。また、apimachineryは広く使われているライブラリ、ということを考えると、この仕様が将来変わるとは…なかなか思えないですね…

どちらかというと、ユーザの設定値のちょっとした違いや、CKEの実装変更により`empty`と`nil`が変わってしまうことを警戒すべきだと思います。

今回やるべきことは

1. CKEのドキュメントに仕様を明記する
2. `empty`と`nil`のテストケースを追加し、CKEの実装変更による壊れを検出できるようにする(万が一apimachineryの仕様変更があったとしても検出できる)

だと思いますがどうでしょう？

docs/cluster.md
```diff
@@ -73,18 +73,19 @@ Taint
...
+| `command_retries`            | false    | *int                             | Number of reboot retries, not including initial attempt. Default: 0     |
+| `command_interval`           | false    | *int                             | Interval of time between reboot retries in seconds. Default: 0          |
+| `evict_retries`              | false    | *int                             | Number of eviction retries, not including initial attempt. Default: 0   |
+| `evict_interval`             | false    | *int                             | Interval of time between eviction retries in seconds. Default: 0        |
+| `max_concurrent_reboots`     | false    | *int                             | Maximum number of nodes to be rebooted concurrently. Default: 1         |
+| `protected_namespaces`       | false    | [`LabelSelector`][LabelSelector] | A label selector to protect namespaces.                                 |
+| `deletable_job_pod_selector` | false    | [`LabelSelector`][LabelSelector] | A label selector to allow deletion of Job-managed Pods.                 |
```
masa213f: `deletable_job_pod_selector` is able to do what we wanted, but I'm concerned about the lack of consistency with `protected_namespaces`.

How about setting a label for JobPods that shouldn't be deleted, similar to `protected_namespaces`?
And If we name the field something like `protected_jobpods`, it would also keep the naming consistent.

I think if we set it up as follows, it will meet our requirements.

```yaml
  protected_jobpods:
    matchExpressions:
    - key: "cke.cybozu.com/job-ok-to-delete"
      operator: DoesNotExist
```

masa213f (changes requested):

docs/cluster.md
YZ775: I aggree with the idea. I'll fix the code.

op/reboot_decide.go
YZ775: I added test cases for `empty` and `nil` in `LabelSelectorAsSelector`.

bin/env
```diff
@@ -1,5 +1,5 @@
 PROJECT=neco-test
-ZONE=asia-northeast1-c
```
masa213f, line 2: How about using GitHub Actions variables here?
https://docs.github.com/ja/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables

masa213f:
Sorry. I'm still reviewing it.

docs/cluster.md
```diff
@@ -104,17 +105,24 @@ The Pods in the non-protected namespaces are also tried to be deleted gracefully
 
 If `protected_namespaces` is not given, all namespaces are protected.
 
+CKE deletes running Job-managed Pods that do **not** match `protected_job_pods`.
+If running Job-managed Pods matching `protected_job_pods` remain, CKE backs off the reboot until the Pods finish on its own.
+
+If `protected_job_pods` is `null` (not given), no Job-managed Pod matches the selector, so all Job-managed Pods are deletable by reboot.
+If it is set to an empty selector (`{}`), all Job-managed Pods match the selector and are protected from deletion.
```
masa213f: Thank you for reflecting the comments.

Looking at this change, I'd like to propose that,

- When `protected_job_pods` is omitted, its default value should be `{}` (i.e., protect all Jobs)

for the following reasons:

- To keep functional consistency with protected_namespaces
    - > If `protected_namespaces` is not given, all namespaces are protected.
- To avoid breaking changes in CKE's behavior.


What do you think?

YZ775: I understand that `protected_namespaces` protects all NS when its is nil because this LabelSelector is used as `selector.String()`.

https://github.com/cybozu-go/cke/blob/b0d4070506f2016ad671cc66cdb20233d4b97470/op/reboot.go#L671

I think it's better to protect all Job Pods when `protected_job_pods` is nil too.
(despite it's reveserd behavaior of `labels.Selector.Matches()`)

bin/env
YZ775: Change the code to use GitHub Actions variables

masa213f (approved):
LGTM

PR #911, opened 2026-08-12
Issue client certificates for cke:user:admin by default

guni1192:
`ckecli kubernetes issue` defaulted to the user name `admin`, which CKE itself also uses to reconcile the cluster. Because both identities were the same string, kube-apiserver audit logs could not tell a human operator from CKE's own reconciliation, and an admission webhook had no way to exclude CKE.

This names certificates issued by the command under a `cke:user:` prefix so the two can be told apart. The default becomes `cke:user:admin`, and the command writes a notice to stderr when `--user` is not given.

Cluster permissions are unaffected: they come from the `system:masters` organization, which is unchanged. `UserKubeconfig` names the context `default` rather than deriving it from the user name, so `kubectl --context=default` keeps working.

**This is a breaking change.** The user name embedded in the kubeconfig changes, so `kubectl --user=admin` and anything reading `.users[0].name` will break. Pass `--user=admin` to restore the previous name. Since MAJOR.MINOR tracks Kubernetes (RELEASE.md), this cannot be signaled by a version bump, which is why the notice goes to stderr.

### DIFF
--- a/pkg/ckecli/cmd/kubernetes_issue.go
+++ b/pkg/ckecli/cmd/kubernetes_issue.go
@@ -26,6 +26,7 @@ var kubernetesIssueCmd = &cobra.Command{
 	Long:  `Issue TLS client certificates for k8s user.`,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
+		// TODO: delete this notice after CKE 1.35.5.
 		if !cmd.Flags().Changed("user") {
 			// stderr, because stdout carries the kubeconfig.
 			fmt.Fprintf(os.Stderr,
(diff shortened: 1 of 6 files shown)


pkg/ckecli/cmd/kubernetes_issue.go
```diff
@@ -25,6 +26,14 @@ var kubernetesIssueCmd = &cobra.Command{
 	Long:  `Issue TLS client certificates for k8s user.`,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if !cmd.Flags().Changed("user") {
+			// stderr, because stdout carries the kubeconfig.
```
YZ775, line 31: Could you add comments like `// delete this warning after CKE 1.35.5` ?

guni1192: thx, fixed: https://github.com/cybozu-go/cke/pull/911/commits/eb60abb7f43b5919bf3db382080e3435e3c9fc2c

YZ775 (approved):
LGTM

PR #913, opened 2026-08-13
stop etcd compaction itself

guni1192:
Both etcd and kube-apiserver have been compacting the keyspace every 5 minutes.
This disables etcd's auto-compaction so that only kube-apiserver compacts.

ref:
- https://github.com/cybozu-private/neco-tasks/issues/963

## Review points

- **Data corruption detection is unaffected.** KV hashes are stored in
  `store.compact()`, which kube-apiserver's Compact RPC reaches through the same
  raft-applied path as auto-compaction, and `InitialCorruptCheck` /
  `--corrupt-check-time` do not involve compaction at all.
- **Existing clusters need a manual grant first.** etcd permits the Compact API
  only for users having the `root` role and CKE grants it only at etcd bootstrap,
  so clusters bootstrapped by CKE older than v1.35.0 stop compacting entirely
  after this change; it is announced as an upgrade note in `CHANGELOG.md`, so
  please check the wording. (Ours: neco-ops#5583, neco-ops#5585)
- **Rollout.** The parameter change makes all etcd members outdated, so
  `EtcdRestartOp` restarts them one by one behind the `EtcdIsGood()` guard.
- **Tests.** Unit tests fix the generated flags and the restart-on-drift path,
  and mtest checks that compaction really happens and that the corruption
  checkers keep running; no test injects real data corruption, which is left to
  etcd's own e2e tests.

depends on
- https://github.com/cybozu-private/neco-ops/issues/5583
- https://github.com/cybozu-private/neco-ops/issues/5585


mtest/etcd_test.go
```diff
@@ -0,0 +1,96 @@
...
+		Expect(err).NotTo(HaveOccurred(), "stdout=%s, stderr=%s", stdout, stderr)
+		Expect(string(stdout)).To(MatchRegexp(`(?m)^Roles:.*\broot\b`), "stdout=%s", stdout)
+	})
+
+	It("should keep checking data corruption periodically", func() {
+		// etcd logs this at startup when --corrupt-check-time is given.  Note that
+		// the initial corruption check leaves no log here, because etcd skips it for
+		// a newly created member.
+		for _, node := range []string{node1, node2, node3} {
+			Expect(etcdLogHas(node, "enabled corruption checking")).To(BeTrue(), "node=%s", node)
```
zeroalphat: I don't think it's necessary to verify in tests whether a specific etcd log is emitted. 
Such tests tend to break when etcd is upgraded, and the cost of keeping up with those changes is high. 
So I'd suggest dropping it.

CHANGELOG.md
```diff
@@ -8,8 +8,26 @@ This project employs a versioning scheme described in [RELEASE.md](RELEASE.md#ve
 
 ### Changed
 
+- Disable etcd auto-compaction and let kube-apiserver do compaction
```
zeroalphat: The CKE changelog follows the style below, so could you write it in the same format?
```
- [PR title] in PR number
```

```suggestion
- stop etcd compaction itself in [#913](https://github.com/cybozu-go/cke/pull/913)
```

mtest/etcd_test.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	})
+
+	It("should compact revisions and check compaction hashes", func() {
+		Eventually(func() error {
```
zeroalphat: This is just a matter of personal preference, but I prefer writing it with a special case polling function.
Since it's a preference, I'll leave it up to you whether to change it.

https://github.com/cybozu-go/cke/blob/e61d070455322e9f0440b29f749cafd88c70e7d7/mtest/repair_test.go#L414-L420
https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

zeroalphat (changes requested):

mtest/etcd_test.go
guni1192: Agreed, dropped all log-based assertions. https://github.com/cybozu-go/cke/pull/913/commits/d46bc201f5d0ab6464f78fb4d5e5d448a687a248

zeroalphat (approved):
LGTM

PR #915, opened 2026-08-13
Remove well.Go from ckecli commands

masa213f:
This is the first step toward removing the dependency on cybozu-go/well.

This PR stops using `well.Go` in ckecli, which was mainly used for signal handling.
Signal handling is now done once in `Execute()`, and subcommands use its context instead.

### DIFF
--- a/pkg/ckecli/cmd/reboot_queue_add.go
+++ b/pkg/ckecli/cmd/reboot_queue_add.go
@@ -1,13 +1,11 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"os"
 	"strings"
 
-	"github.com/cybozu-go/well"
 	"github.com/spf13/cobra"
 
 	"github.com/cybozu-go/cke"
@@ -22,6 +20,8 @@ The nodes should be specified with their IP addresses.
 If FILE is -, the contents are read from stdin.`,
 	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := cmd.Context()
+
 		f := os.Stdin
 		if args[0] != "-" {
 			var err error
@@ -36,29 +36,25 @@ If FILE is -, the contents are read from stdin.`,
 		if err != nil {
 			return err
 		}
-		nodes := strings.Fields(string(data))
+		nodes := strings.FieldsSeq(string(data))
 
-		well.Go(func(ctx context.Context) error {
-			for _, node := range nodes {
-				entry := cke.NewRebootQueueEntry(node)
-				cluster, err := storage.GetCluster(ctx)
-				if err != nil {
-					return err
-				}
-				err = validateNode(node, cluster)
-				if err != nil {
-					return err
-				}
-
-				err = storage.RegisterRebootsEntry(ctx, entry)
-				if err != nil {
-					return err
-				}
+		for node := range nodes {
+			entry := cke.NewRebootQueueEntry(node)
+			cluster, err := storage.GetCluster(ctx)
+			if err != nil {
+				return err
 			}
-			return nil
-		})
-		well.Stop()
-		return well.Wait()
+			err = validateNode(node, cluster)
+			if err != nil {
+				return err
+			}
+
+			err = storage.RegisterRebootsEntry(ctx, entry)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
 	},
 }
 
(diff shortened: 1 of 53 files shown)


pkg/ckecli/cmd/reboot_queue_add.go
```diff
@@ -36,29 +36,25 @@ If FILE is -, the contents are read from stdin.`,
 		if err != nil {
 			return err
 		}
-		nodes := strings.Fields(string(data))
```
masa213f, line 39: I fixed the lint errors.
```
Run make lint
go tool golangci-lint run
Error: pkg/ckecli/cmd/reboot_queue_add.go:39:12: stringsseq: Ranging over FieldsSeq is more efficient (modernize)
		nodes := strings.Fields(string(data))
		         ^
1 issues:
* modernize: 1
make: *** [Makefile:26: lint] Error 1
Error: Process completed with exit code 2.
```

pkg/ckecli/cmd/ssh.go
```diff
@@ -222,11 +221,7 @@ If COMMAND is specified, it will be executed on the node.
 
 	Args: cobra.MinimumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		well.Go(func(ctx context.Context) error {
-			return sshSubMain(ctx, args)
-		})
-		well.Stop()
-		return well.Wait()
+		return sshSubMain(cmd.Context(), args)
```
YZ775, line 248: ssh agent may leak when command recieves signals because `exec.CommandContext()` does not execute command when ctx is already canceled.

This is not happend when well is used because there are 5 seconds delay before canceling ctx.

So, it's better to use another context

https://github.com/cybozu-go/cke/blob/9aaf0cd40d693d01b73779f202b328796bb756a3/pkg/ckecli/cmd/ssh.go#L134
```go
func killSshAgent(ctx context.Context) error {
	ctx = context.WithoutCancel(ctx)

	cmd := exec.CommandContext(ctx, "ssh-agent", "-k")
}

File: pkg/ckecli/cmd/root.go [hunk 3]
```diff
@@ -79,9 +83,34 @@ with etcd.  CKE server watches etcd to receive any updates.`,
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	if err := rootCmd.Execute(); err != nil {
-		log.ErrorExit(err)
+	sigCh := make(chan os.Signal, 1)
```
YZ775: This can be simplify be using `signal.NotifyContext`

```go
func Execute() {
	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
	defer stop()

	err := rootCmd.ExecuteContext(ctx)
	if err == nil && ctx.Err() != nil {
		err = errors.New("interrupted")
	}
	if err != nil {
		log.ErrorExit(err)
	}
}
```

YZ775 (changes requested):

pkg/ckecli/cmd/root.go
masa213f: I was intentionally using `signal.Notify` to log the received signals.
However... since there's little need to log signals in the command-line interface, so I'll switch to `signal.NotifyContext` as you pointed out.

pkg/ckecli/cmd/ssh.go
masa213f: That's ture! fixed. 666700e

There is another bug, so I fixed it as well. 84c530b
If the context is canceled before ssh-agent starts, the opening FIFO may hang.

YZ775 (approved):
LGTM

pkg/ckecli/cmd/scp.go
```diff
@@ -62,9 +61,9 @@ func scpSubMain(ctx context.Context, args []string) error {
 			})
 		}
 	}()
-	defer func() { _ = killSshAgent(ctx) }()
+	defer killSshAgent()
 
-	if err = writeToFifo(pipeFilename, pirvateKey); err != nil {
+	if err = writeToFifo(ctx, pipeFilename, pirvateKey); err != nil {
```
YZ775: Sorry, I found typo. Could you fix this too?

```suggestion
	if err = writeToFifo(ctx, pipeFilename, privateKey); err != nil {
```

YZ775 (changes requested):

YZ775 (approved):
Thank you!

PR #919, opened 2026-08-25
rivers: Drop cybozu-go/log and cybozu-go/well dependencies

masa213f:
This PR removes the `github.com/cybozu-go/log` and `github.com/cybozu-go/well` dependencies from `tools/rivers`, replacing them with the standard libraries.

While rewriting the signal/shutdown handling that `well` used to provide, I also took the opportunity to clean up rivers' internal structure a bit.

### DIFF
--- a/tools/rivers/main.go
+++ b/tools/rivers/main.go
@@ -35,10 +35,7 @@ func main() {
 	upstreamAddrs := strings.Split(*flgUpstreams, ",")
 	upstreams := make([]*Upstream, len(upstreamAddrs))
 	for i, a := range upstreamAddrs {
-		upstreams[i] = &Upstream{
-			address: a,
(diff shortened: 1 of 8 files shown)


tools/rivers/main.go
morimoto-cybozu: It would be good for `upstream.go` to provide `NewUpstream(address string)`.
`main.go` should not take care of `conns`.

Or we can leave `conns` `nil`.

```suggestion
upstreams[i] = &Upstream{address: a}
```

morimoto-cybozu (changes requested):

tools/rivers/main.go
masa213f: That's good. I've made the change.

morimoto-cybozu (approved):

